PR #4, opened 2020-03-09
Ml production

MischaPanch:


src/sensai/featuregen.py
```diff
@@ -269,6 +278,17 @@ def _generate(self, df: pd.DataFrame, ctx=None) -> pd.DataFrame:
         return resultDf
 
 
+def getFlattenedFeatureGenerator(fgen: FeatureGenerator, ruleTemplate: data_transformation.DFTNormalisation.RuleTemplate):
```
opcode81: This function is too specific. 
Why can I only apply it to flatten all columns?
Why can I only use a rule template?

Also, the prefix "get" in its name is not appropriate here.

src/sensai/featuregen.py
```diff
@@ -269,6 +278,17 @@ def _generate(self, df: pd.DataFrame, ctx=None) -> pd.DataFrame:
...
+    """
+    Return a feature generator that will generate flattened output of the provided fgen
+
+    :param fgen: feature generator to be flattened
+    :param ruleTemplate: ruleTemplate to apply to all output columns
```
opcode81: The parameter should always be called "normalisationRuleTemplate" to make its purpose clear.

src/sensai/featuregen.py
```diff
@@ -243,21 +243,30 @@ class FeatureGeneratorFlattenColumns(RuleBasedFeatureGenerator):
...
     will be created. It will contain the columns "vec1_<i1>", "vec2_<i2>" with i1, i2 ranging in (0, dim1), (0, dim2).
 
     """
-    def __init__(self, columns: Union[str, Sequence[str]], categoricalFeatureNames: Sequence[str] = (),
+    def __init__(self, columns: Optional[Union[str, Sequence[str]]] = None, categoricalFeatureNames: Sequence[str] = (),
             normalisationRules: Sequence[data_transformation.DFTNormalisation.Rule] = (),
             generatedColumnsRuleTemplate: data_transformation.DFTNormalisation.RuleTemplate = None):
+        """
+
+        :param columns: column name, list of columns or None. If None, all columns in the input dataframe fill be flattened
```
opcode81: name of the column or list of names of columns to be flattened. If None, all columns will be flattened.
It's good to have the phrase "to be flattened" in there...

src/sensai/featuregen.py
```diff
@@ -243,21 +243,30 @@ class FeatureGeneratorFlattenColumns(RuleBasedFeatureGenerator):
     Instances of this class take columns with vectors and creates a dataframe with columns containing entries of
     these vectors.
 
-    For example, if columns "vec1", "vec2" contain vectors of dimensions dim1, dim2, a datafrane dim1+dim2 new columns
+    For example, if columns "vec1", "vec2" contain vectors of dimensions dim1, dim2, a dataframe with dim1+dim2 new columns
```
opcode81: If we're correcting this, let's go all the way: "data frame"

src/sensai/featuregen.py
```diff
@@ -24,28 +24,45 @@ class FeatureGenerator(ABC):
...
+                 normalisationRules: Sequence[data_transformation.DFTNormalisation.Rule] = (),
+                 generatedColumnsRuleTemplate: data_transformation.DFTNormalisation.RuleTemplate = None, addCategoricalDefaultRules=True):
         """
         :param categoricalFeatureNames: if provided, will ensure that the respective columns in the generated data frames will
             have dtype 'category'.
             Furthermore, presence of meta-information can later be leveraged for further transformations, e.g. one-hot encoding.
         :param normalisationRules: Rules to be used by DFTNormalisation (e.g. for constructing an input transformer for a model).
             These rules are only relevant if a DFTNormalisation object consuming them is instantiated and used
             within a data processing pipeline. They do not affect feature generation.
+        :param generatedColumnsRuleTemplate: If given, a normalisation rule based on this template will be created for all generated columns.
```
opcode81: Parameter should be called "normalisationRuleTemplate".
"This parameter can be supplied instead of normalisationRules for the case where there shall be a single rule that applies to all columns generated by this feature generator."

src/sensai/featuregen.py
```diff
@@ -24,28 +24,45 @@ class FeatureGenerator(ABC):
...
+
+    def _inputValidation(self):
+        if self.generatedColumnsRuleTemplate is not None:
+            if self.getNormalisationRules():
+                raise ValueError(f"normalisationRules should be empty when a generatedColumnsRuleTemplate is provided, instead got {self.getNormalisationRules()}")
+
     def getNormalisationRules(self) -> Sequence[data_transformation.DFTNormalisation.Rule]:
         return self._normalisationRules
 
+    def getGeneratedColumnsRule(self):
```
opcode81: Why do we need this?

src/sensai/featuregen.py
```diff
@@ -24,28 +24,45 @@ class FeatureGenerator(ABC):
...
         self._categoricalFeatureNames = categoricalFeatureNames
         normalisationRules = list(normalisationRules)
         if addCategoricalDefaultRules and len(categoricalFeatureNames) > 0:
             normalisationRules.append(data_transformation.DFTNormalisation.Rule(r"(%s)" % "|".join(categoricalFeatureNames), unsupported=True))
             normalisationRules.append(data_transformation.DFTNormalisation.Rule(r"(%s)_\d+" % "|".join(categoricalFeatureNames), skip=True))
         self._normalisationRules = normalisationRules
 
+        self._inputValidation()
+
+    def _inputValidation(self):
```
opcode81: This should be done directly in the constructor and not based on the members but the actual parameters. Make it the first thing the constructor does.

src/sensai/featuregen.py
```diff
@@ -86,6 +103,11 @@ def generate(self, df: pd.DataFrame, ctx=None) -> pd.DataFrame:
                 if series.dtype.name != 'category':
                     resultDF[colName] = series.astype('category', copy=False)
 
+        self.generatedColumnNames = resultDF.columns
+        if self.generatedColumnsRuleTemplate is not None:
+            self._generatedColumnsRule = data_transformation.DFTNormalisation.Rule(r"(%s)" % "|".join(self.generatedColumnNames), **self.generatedColumnsRuleTemplate.__dict__)
```
opcode81: The names must be escaped with re.escape

src/sensai/featuregen.py
```diff
@@ -195,44 +218,60 @@ def _generateFeatureDict(self, namedTuple) -> Dict[str, Any]:
...
     def _generate(self, df: pd.DataFrame, ctx=None) -> pd.DataFrame:
+        if self.columns is None:
+            return df
         missingCols = set(self.columns) - set(df.columns)
         if len(missingCols) > 0:
             raise Exception(f"Columns {missingCols} not present in data frame; available columns: {list(df.columns)}")
         return df[self.columns]
 
 
-class FeatureGeneratorTakeAllColumns(RuleBasedFeatureGenerator):
```
opcode81: This deletion must be reverted.

src/sensai/featuregen.py
opcode81: You are not actually checking the second condition you mentioned in the docstring: no categorical features. Then again, we don't necessarily need to. We could make the template apply to all non-categorical features only; that would be the most flexible way of handling things.

src/sensai/featuregen.py
MischaPanch: We don't really need it. I thought it might be useful as info for the user (to see what rule a generator has applied once it ran) but probably it is more confusing than helping. I'll remove it and add a getter for the ruleTemplate instead

src/sensai/featuregen.py
MischaPanch: By checking after all normalisation rules have been set, I also check against rules for categorical features, right? That's why I did it last and based on the members.

Applying the template to non-categorical features only makes the template unusable for declaring normalisation categorical input as unsupported, as would be needed by featuregen from columngen and possibly others. Nevertheless, we can consider applying rule templates only to non-categorical features and handling the categorical ones separately

src/sensai/featuregen.py
MischaPanch: Ok, I'll mark it as deprecated

opcode81: Why? There is no reason to deprecate it.

src/sensai/featuregen.py
opcode81: Categorical features are always unsupported; that's what the default rule that is auto-generated states.

opcode81: Anyway, since auto-generation of rules for categorical features always makes sense, the rule template should not apply to them and there isn't really a reason why a rule template should not be usable for the remaining columns in such cases.

src/sensai/featuregen.py
```diff
@@ -24,28 +24,45 @@ class FeatureGenerator(ABC):
...
             These rules are only relevant if a DFTNormalisation object consuming them is instantiated and used
             within a data processing pipeline. They do not affect feature generation.
+        :param generatedColumnsRuleTemplate: If given, a normalisation rule based on this template will be created for all generated columns.
+            As with other rules, this is only relevant if DFTNormalization is used after feature generation.
+            The rule template should only be provided if no categoricalFeatureNames and normalisationRules have been passed
         :param addCategoricalDefaultRules:
             If True, normalisation rules for categorical features (which are unsupported by normalisation) and their corresponding one-hot
             encoded features (with "_<index>" appended) will be added.
         """
+        self.generatedColumnNames = None
```
opcode81: We should make clear that it's the list of column names that was generated by the most recent generation that this object performed. To make this clear we at least need to rename this column; even more clarity can be achieved by using a getter instead and making this attribute private, such that we can add the necessary documentation.

src/sensai/featuregen.py
```diff
@@ -86,6 +103,11 @@ def generate(self, df: pd.DataFrame, ctx=None) -> pd.DataFrame:
                 if series.dtype.name != 'category':
                     resultDF[colName] = series.astype('category', copy=False)
 
+        self.generatedColumnNames = resultDF.columns
+        if self.generatedColumnsRuleTemplate is not None:
+            self._generatedColumnsRule = data_transformation.DFTNormalisation.Rule(r"(%s)" % "|".join(self.generatedColumnNames), **self.generatedColumnsRuleTemplate.__dict__)
+            self._normalisationRules.append(self._generatedColumnsRule)
```
opcode81: Is anything preventing this from being executed more than once?

src/sensai/featuregen.py
MischaPanch: Done

src/sensai/featuregen.py
MischaPanch: Done

src/sensai/featuregen.py
MischaPanch: Done

src/sensai/featuregen.py
MischaPanch: I adjusted the mechanism of the rule template such that it only applies to non-categorical features

src/sensai/featuregen.py
MischaPanch: As discussed, removed in favor of default behavior of TakeCollumnsFeatureGenerator

src/sensai/featuregen.py
MischaPanch: Done

src/sensai/featuregen.py
MischaPanch: Whoops, that was a rather big mistake. I have fixed it by separating rules for categorical features and for the rest

src/sensai/featuregen.py
MischaPanch: Done. This was also relevant in other places, I have written a helper function in util.string

src/sensai/featuregen.py
MischaPanch: Done

src/sensai/featuregen.py
MischaPanch: I extended the function. This extension necessitated a new feature in TakeColumns generator which I have implemented in the same commit

src/sensai/featuregen.py
```diff
@@ -195,44 +221,62 @@ def _generateFeatureDict(self, namedTuple) -> Dict[str, Any]:
...
+        missingCols = columnsToTake.difference(df.columns)
         if len(missingCols) > 0:
             raise Exception(f"Columns {missingCols} not present in data frame; available columns: {list(df.columns)}")
-        return df[self.columns]
-
 
-class FeatureGeneratorTakeAllColumns(RuleBasedFeatureGenerator):
-    def _generate(self, df: pd.DataFrame, ctx=None) -> pd.DataFrame:
-        return df
+        return df[sorted(columnsToTake)]
```
opcode81: Typically, column order should not matter, but maybe it would be better if we didn't override the user's ordering given in self.columns (for the case where he specifies it). We can filter out exceptColumns with a list comprehension instead of using sets.

src/sensai/featuregen.py
```diff
@@ -542,3 +599,36 @@ def _generate(self, df: pd.DataFrame, ctx=None) -> pd.DataFrame:
         else:
             _log.info(f"Generating target features via {self.vectorModel.__class__.__name__}")
             return self.vectorModel.predict(df)
+
+
+def flattenFeatureGenerator(fgen: FeatureGenerator, columnsToFlatten: List[str] = None,
```
opcode81: I would propose the name "flattenedFeatureGenerator", as it makes clear that it returns a new one and does not (somehow) flatten the existing one.
Perhaps even "FlattenedFeatureGenerator", as it's akin to a construction of a new instance (though some might disagree with the notion of naming a function this way in general).

src/sensai/featuregen.py
```diff
@@ -195,44 +221,62 @@ def _generateFeatureDict(self, namedTuple) -> Dict[str, Any]:
 
 
 class FeatureGeneratorTakeColumns(RuleBasedFeatureGenerator):
-    def __init__(self, columns: Union[str, List[str]], categoricalFeatureNames: Sequence[str] = (),
-            normalisationRules: Sequence[data_transformation.DFTNormalisation.Rule] = ()):
-        super().__init__(categoricalFeatureNames=categoricalFeatureNames, normalisationRules=normalisationRules)
+    def __init__(self, columns: Union[str, List[str]] = None, exceptColumns: List[str] = None, categoricalFeatureNames: Sequence[str] = (),
```
opcode81: Note: We could use Sequence[str] and default () here; it would simplify the definition below.
(It would also establish a difference to columns where None has an entirely different interpretation.)

src/sensai/featuregen.py
MischaPanch: Done

src/sensai/featuregen.py
MischaPanch: Good point, I adjusted it

src/sensai/featuregen.py
MischaPanch: Done. Aaand, I'm merging this :)

PR #5, opened 2020-03-11
Feature/setup on windows

schroedk:


requirements.txt
```diff
@@ -2,10 +2,8 @@ pandas==0.25.3
 scipy==1.4.1
 numpy==1.18.1
 scikit-learn==0.22.1
-pysqlite3==0.4.1
```
opcode81: Don't we still need pysqlite3?

test/test_dummy_sklearn.py
```diff
@@ -0,0 +1,3 @@
+def test_dummy_sklearn():
```
MischaPanch: You could include imports of the relevant libraries here, thereby testing that they get installed in the testenv during the build (and remove the dummy from the modules name)

tox.ini
```diff
@@ -4,18 +4,30 @@
...
-envlist = py37, docs
+envlist = py, docs
 isolated_build = True
 
 [testenv]
 extras =
     test
 commands =
-    pytest
+    pytest test/test_dummy.py -v
```
MischaPanch: We should not test single files but rather all files within a directory. This means that the test directory should get subdirs like
test
       base
       torch
       tensorflow

.github/workflows/tox.yaml
```diff
@@ -25,12 +25,12 @@ jobs:
       - name: Install Tox and any other packages
         run: pip install tox
       - name: Run Tox
-        run: tox -e py -e docs # Run tox using the version of Python in `PATH`
+        run: tox -e py -e torch -e tensorFlow -e docs # Run tox using the version of Python in `PATH`
```
MischaPanch: this could be shortened to
```
tox -e py,torch,tensorflow,docs
```
I just found out about this syntax :)

requirements.txt
schroedk: Is there a reason, we need

https://github.com/coleifer/pysqlite3

as we are only using build-in sqlite3 in cache.py?

opcode81: Good catch. Right you are.

opcode81:
Can be merged as far as I'm concerned

.github/workflows/tox.yaml
schroedk: Done

tox.ini
schroedk: Done

test/test_dummy_sklearn.py
schroedk: Done

requirements.txt
schroedk: Done

.github/workflows/tox.yaml
schroedk: Actually you have to write
tox -e py, torch, tensorflow, -e docs

as otherwise docs will not be build.

PR #6, opened 2020-03-11
Feature/make lightgbm optional

schroedk:


opcode81:
This PR includes the changes of another PR and therefore can be addressed only once the other has been handled.

src/sensai/lightgbm/lightgbm.py
```diff
@@ -0,0 +1,60 @@
+from typing import Sequence
+import logging
+import lightgbm
+import pandas as pd
+
+from sensai.sklearn.sklearn_base import AbstractSkLearnMultipleOneDimVectorRegressionModel, \
```
opcode81: Never use absolute imports within the library.

schroedk: Done

PR #7, opened 2020-04-29
Simplifications in hyperopt

MischaPanch:
I am integrating some hyperopt stuff in the project I'm currently working on (clustering) and wanted to propose some simplifications in the module while doing that :)

### DIFF
--- a/src/sensai/evaluation.py
+++ b/src/sensai/evaluation.py
@@ -13,19 +13,20 @@ import matplotlib.pyplot as plt
 
 from .util.io import ResultWriter
 from .util.typing import PandasNamedTuple
-from .vector_model import InputOutputData, VectorModel, PredictorModel, VectorClassificationModel, VectorRegressionModel
+from .vector_model import InputOutputData, VectorModel, PredictorModel, VectorClassificationModel, VectorRegressionModel, \
+    VectorModel
 from .eval_stats import RegressionEvalStats, ClassificationEvalStats, RegressionEvalStatsCollection, \
-    ClassificationEvalStatsCollection
+    ClassificationEvalStatsCollection, EvalStats, EvalStatsCollection
 
 
 _log = logging.getLogger(__name__)
 
-TModel = TypeVar("TModel", VectorRegressionModel, VectorClassificationModel)
-TEvalStats = TypeVar("TEvalStats", RegressionEvalStats, ClassificationEvalStats)
-TEvalStatsCollection = TypeVar("TEvalStatsCollection", RegressionEvalStatsCollection, ClassificationEvalStatsCollection)
-TEvaluator = TypeVar("TEvaluator", "VectorRegressionModelEvaluator", "VectorClassificationModelEvaluator")
-TEvalData = TypeVar("TEvalData", "VectorRegressionModelEvaluationData", "VectorClassificationModelEvaluationData")
-TCrossValData = TypeVar("TCrossValData", "VectorClassificationModelCrossValidationData", "VectorRegressionModelCrossValidationData")
+TModel = TypeVar("TModel", bound=VectorModel)
+TEvalStats = TypeVar("TEvalStats", bound=EvalStats)
+TEvalStatsCollection = TypeVar("TEvalStatsCollection", bound=EvalStatsCollection)
+TEvaluator = TypeVar("TEvaluator", bound="VectorModelEvaluator")
+TEvalData = TypeVar("TEvalData", bound="VectorModelEvaluationData")
+TCrossValData = TypeVar("TCrossValData", bound="VectorModelCrossValidationData")
 
 
 class VectorModelEvaluationData(ABC, Generic[TEvalStats]):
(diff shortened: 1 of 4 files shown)


src/sensai/hyperopt.py
```diff
@@ -152,49 +152,32 @@ def _saveCSV(self):
...
     _log = _log.getChild(__qualname__)
 
-    def __init__(self, modelFactory: Callable[..., VectorModel], parameterOptions: Union[Dict[str, Sequence[Any]], List[Dict[str, Sequence[Any]]]],
+    def __init__(self, modelFactory: Callable[..., VectorModel], parameterOptions: Dict[str, Sequence],
             numProcesses=1, csvResultsPath=None, parameterCombinationSkipDecider: ParameterCombinationSkipDecider = None):
         """
         :param modelFactory: the function to call with keyword arguments reflecting the parameters to try in order to obtain a model instance
-        :param parameterOptions: a dictionary which maps from parameter names to lists of possible values - or a list of such dictionaries,
-            where each dictionary in the list has the same keys
+        :param parameterOptions: a dictionary which maps from parameter names to lists of possible values
```
opcode81, line 166: Why would you remove the option to provide a list? This reduces functionality.

src/sensai/hyperopt.py
```diff
@@ -236,19 +219,13 @@ def collectResult(values):
...
-        else:
-            executor = ProcessPoolExecutor(max_workers=self.numProcesses)
-            futures = []
-            for parameterOptions in self.parameterOptionsList:
-                for i, paramsDict in enumerate(iterParamCombinations(parameterOptions)):
-                    futures.append(executor.submit(self._evalParams, self.modelFactory, evaluatorOrValidator, self.parameterCombinationSkipDecider,
-                        **paramsDict))
-            for i, future in enumerate(futures):
-                collectResult(future.result())
+        executor = ProcessPoolExecutor(max_workers=self.numProcesses)
```
opcode81, line 222: Why would you remove the option to not use multi-processing?

src/sensai/hyperopt.py
opcode81: You stated: "It is probably no too much to ask a user to call update if they have a list of dicts already"
Update will not achieve what the list supports - to the contrary. Note that the list must contain dictionaries where the keys are already all the same, so update would give you the last dictionary only.
The purpose of the list is to avoid full combinatorics. For example, if it only makes sense to explore activation=sigmoid in combination with targetTransformer=maxAbsScaler and activation=none in combination with targetTransformer=standardScaler, then you can have two dictionaries which otherwise contain the same parameter options for any further parameters.

src/sensai/hyperopt.py
opcode81: You stated: "It increases the code complexity and I can't see any benefit from it"
There is an obvious benefit to not using multi-processing: No need to send anything across (via pickling) to another process. It is much better to use the main process to do everything.

src/sensai/evaluation.py
```diff
@@ -344,12 +344,11 @@ def _createResultData(self, trainedModels, evalDataList, testIndicesList, predic
 
 def computeEvaluationMetricsDict(model, evaluatorOrValidator: Union[VectorModelEvaluator, VectorModelCrossValidator]) -> Dict[str, float]:
     if isinstance(evaluatorOrValidator, VectorModelEvaluator):
-        evaluator: VectorModelEvaluator = evaluatorOrValidator
```
opcode81, line 346: Please avoid purely cosmetic changes like this (whether it's an improvement is debatable). We don't need to have changesets that complicate the history for things that aren't really an improvement.

src/sensai/hyperopt.py
MischaPanch: I misunderstood the purpose of avoiding full combinatorics, I'll revert it

src/sensai/hyperopt.py
MischaPanch: I guess I underestimated the overhead of multiprocessing. You are right, it's not worth it just for getting rid of some lines of code. Since the only remaining changes are cosmetic, I'm closing this PR

src/sensai/evaluation.py
MischaPanch: It's gone

src/sensai/hyperopt.py
```diff
@@ -152,17 +162,15 @@ def _saveCSV(self):
...
-        values["str(model)"] = str(model)
-        self.addValues(values)
-
     def getDataFrame(self) -> pd.DataFrame:
         return self.df
 
 
 class GridSearch(TrackedExperimentDataProvider):
+    """
+    Instances of this class can be used for evaluating clustering models with different user-provided parametrizations
```
opcode81, line 171: This class is for all types of models not only clustering models.

MischaPanch: Whoops, classical copy paste error

PR #9, opened 2020-07-26
Refactoring/splitting up evaluation

MischaPanch:
This PR contains a large restructuring and no new features. I realize that reviewing it will be quite some work. Please bear with me, I really think it is necessary.

I have written some of the reasons in a commit message but here some more:

1) It is a first step towards enabling us to have a common class that are less specific than VectorModel. It is not the last though - local search and hyperopt should work for the more general models too. I will refactor the corresponding modules once this PR is through
2) Even for me the large modules that I now split in multiple parts were hard to read. I think the structure as it is now will alleviate contributions by others

I'm sorry that most of it is in one large commit but I simply saw no good way to split up the work in different ones. Note that even though much is highlighted as new code, it really mostly is the existing codebase adjusted to the new structure. From a functional point of view, the only real changes are new common abstractions for models and evaluators

The PR is WIP for now as I will add more tests to make sure that the refactoring has not broken anything. But it should be mature enough for you to be able to review productively


src/sensai/evaluation/eval_stats.py
```diff
@@ -0,0 +1,122 @@
...
+import seaborn as sns
+from matplotlib import pyplot as plt
+
+from ..models.base import FitterModel, PredictorModel
+
+TEvalStats = TypeVar("TEvalStats", bound='EvalStats')
+TMetric = TypeVar("TMetric", bound="Metric")
+
+
+# TODO or not TODO: the inheritance structure here creates a circular dependency of eval stats and metrics.
```
MischaPanch, line 15: Out of the many TODOs that I have added, I think this one deserves particular attention. I was really sad when I realized that I could not separate metrics from eval stats, it blew up the corresponding classification and regression modules. What do you think?

opcode81:
In general, I think it would good practice to discuss bigger refactorings before we actually perform them - unless you have no problem with a rejection afterwards.

MischaPanch:
> In general, I think it would good practice to discuss bigger refactorings before we actually perform them - unless you have no problem with a rejection afterwards.

Rejection is fine :), I knew it was a risk. I just wanted to get the conversation about a more flexible interface started. I'll explain my reasons for this structure in a separate comment

MischaPanch:
So, concerning the longer comment above
* Completely agree on the git-split ansatz
* My main idea was to allow evaluation of classes that are not vector model. The clustering stuff I built has a similar evaluation structure, so it seemed natural that the eval stats, eval data and evaluators themselves should have a common interface. For that we need a common model interface, that's the reason for the very generic abstract classes. 
Then I would go on to have hyperopt and local search to accept the general evaluator classes, because these modules can be used for all kinds of evaluations (not just vector model). This certainly can be achieved with a different grouping as well

MischaPanch:
In conclusion, I'm closing this PR and will open a new one with a different structure and the git history intact

PR #10, opened 2020-07-31
Refactoring/generalized evaluation

MischaPanch:


src/sensai/evaluation/eval_stats/eval_stats_regression.py
```diff
@@ -0,0 +1,261 @@
...
+    @classmethod
+    def computeValue(cls, y_true: np.ndarray, y_predicted: np.ndarray):
+        return np.median(cls.computeAbsErrors(y_true, y_predicted))
+
+
+class RegressionEvalStats(PredictionEvalStats["RegressionMetric"]):
+    """
+    Collects data for the evaluation of predicted continuous values and computes corresponding metrics
+
+    :param y_predicted: the predicted values
```
opcode81, line 117: The params docstrings are in the wrong place.

src/sensai/evaluation/evaluation.py
```diff
@@ -0,0 +1,213 @@
...
+
+log = logging.getLogger(__name__)
+
+TModel = TypeVar("TModel", bound=VectorModel)
+TEvalStats = TypeVar("TEvalStats", bound=EvalStats)
+TEvalStatsCollection = TypeVar("TEvalStatsCollection", bound=EvalStatsCollection)
+
+
+# TODO: I'm not sure about the name and signature. Should we add a more generic model type and call this ModelEvaluator?
+class MetricsEvaluator(ABC):
```
opcode81, line 23: As originally suggested, I would prefer the name **MetricsComputation**, as the term "evaluator" is already reserved for different semantics. Or perhaps **MetricsDictProvider** or **MetricsDictComputation**?

src/sensai/evaluation/eval_stats/eval_stats_regression.py
MischaPanch: I did it on purpose. When the parameters are above the init, the rendered documentation looks much better (one can exclude the special member __init__ then). Otherwise I think part of the docu can get lost. At the same time pycharm is able to read the parameters from docstrings in this form. So I think there is no real downside to putting docstrings above the init. In some later commit I would adjust all remaining docstrings, if you agree

src/sensai/evaluation/evaluation.py
MischaPanch: +1 for MetricsDictProvider

src/sensai/evaluation/eval_stats/eval_stats_regression.py
opcode81: I do not agree. The rendering of some piece of software should not have priority here.
The logical place to describe the constructor's arguments is in the constructor's dedicated docstring, while the class' docstring should describe what the class represents/does. Mixing that information with constructor-specific information is weird, especially since the class could provide alternative means of construction as well, and the constructor's args are just the "common denominator" (i.e. the most low-level data suitable for construction, typically) and not universally relevant. (I realise that calling the class is equivalent to calling __init__, so its not completely devoid of logic, but the argument still stands.)
Consistency with the rest of the project is also relevant; all other classes use the other style.
FWIW, PEP257 (Docstring Conventions) agrees with my notion: "The docstring for a class should summarize its behavior ... The class constructor should be documented in the docstring for its __init__ method. Individual methods should be documented by their own docstring."

MischaPanch: I'll change the docstring accordingly (directly on master)

PR #11, opened 2020-08-06
Feature/added clustering evaluation

MischaPanch:
The penultimate PR (the last one will just add minor functionality) that adds the clustering evaluation. There is one change relevant to the rest of the library in the commit 679c6cb4 (apart from other minor improvements here and there)

I added a notebook demonstrating the new methods, later on I'll run it as integration test in the CI pipeline. For now, in case you want to try it out, just write me a message. I'll send you the sample data then


src/sensai/evaluation/eval_stats/eval_stats_base.py
```diff
@@ -52,8 +57,7 @@ def __str__(self):
 
 
 class Metric(Generic[TEvalStats], ABC):
-    def __init__(self, name: str):
```
opcode81, line 55: Making the name static reduces generality as we lose the ability to have the name depend on the parametrisation of the metric.
Why did you change this?

src/sensai/evaluation/eval_stats/eval_stats_classification.py
```diff
@@ -54,8 +50,12 @@ def _computeValue(self, y_true, y_predicted, y_predictedClassProbabilities):
 
 
 class ClassificationMetricTopNAccuracy(ClassificationMetric):
+    requiresProbabilities = True
+
     def __init__(self, n: int):
-        super().__init__(f"Top{n}Accuracy", requiresProbabilities=True)
+        # TODO or not TODO: this is the only metric where the name is not static. We could make a static name like
+        #   "TopNAccuracy" without losing much information
```
opcode81, line 57: We would lose _vital_ information. We sometimes use several instances of this with different values of n, and we obviously need to be able to differentiate them when we get our results back.
I don't understand why you would change this in the first place. You are reducing flexibility rather than increasing it (for no apparent reason); this dualism of sometimes static sometimes not is quite strange. There is no reason not to have the string attribute in every instance, really.

src/sensai/evaluation/eval_stats/eval_stats_classification.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, y_predicted: PredictionArray = None,
         :param y_predictedClassProbabilities: a data frame whose columns are the class labels and whose values are probabilities
         :param labels: the list of class labels
         :param metrics: the metrics to compute for evaluation; if None, use default metrics
-        :param additionalMetrics: the metrics to additionally compute
+        :param additionalMetrics: the metrics to additionally compute. This should only be provided if metrics is None
```
opcode81, line 86: The addition is not required as there is no such restriction.

src/sensai/evaluation/eval_stats/eval_stats_regression.py
```diff
@@ -120,18 +108,16 @@ def __init__(self, y_predicted: PredictionArray, y_true: PredictionArray,
         :param y_predicted: the predicted values
         :param y_true: the true values
         :param metrics: the metrics to compute for evaluation; if None, use default metrics
-        :param additionalMetrics: the metrics to additionally compute
+        :param additionalMetrics: the metrics to additionally compute. This should only be provided if metrics is None
```
opcode81, line 111: Same.

src/sensai/evaluation/eval_stats/eval_stats_base.py
```diff
@@ -21,13 +21,17 @@
...
 class EvalStats(Generic[TMetric]):
-    def __init__(self, metrics: List[TMetric]):
+    def __init__(self, metrics: List[TMetric], additionalMetrics: List[TMetric] = None):
         if len(metrics) == 0:
             raise ValueError("No metrics provided")
         self.metrics = metrics
+        # Implementations of EvalStats will typically provide default metrics, therefore we include
+        # the possibility for passing additional metrics here
+        if additionalMetrics is not None:
+            self.metrics.extend(additionalMetrics)
```
opcode81, line 31: It is not safe to just extend metrics, as metrics was passed externally (and therefore it can be quite unexpected for it to be mutated).
My original implementations copied metrics.

src/sensai/evaluation/eval_stats/eval_stats_classification.py
MischaPanch: But it is a valid recommendation for the user. It does not make sense to provide additional metrics when the default ones have been overrode, right?

src/sensai/evaluation/eval_stats/eval_stats_base.py
MischaPanch: Right, I'll add the copying part

src/sensai/evaluation/eval_stats/eval_stats_base.py
MischaPanch: I explained it in the commit. My particular use case was that I needed the metric name for parsing and sorting but only had access to it when instantiating the object. Sure, the metric objects are lioghtweight, but it still felt strange having to instantiate them just for the name

MischaPanch: In any case, I knew that there would be a discussion about it, that's why the change is in one commit and easily revertible. When I changed it I hadn't realized the one exception where the name actually depended on the init parameters

src/sensai/evaluation/eval_stats/eval_stats_classification.py
MischaPanch: See comment above, I agree that the attribute should be present in every instance. I currently don't see an elegant solution for this case (and others similar to it). If we can't come up with one, all is left is to revert this commit

MischaPanch:
Update: instead of opening another PR later on, I added the remaining functionality related to spanning trees in a commit on top of this branch. Thus, this PR concludes the migration of clustering stuff from appliedAI's repo to sensAI

src/sensai/evaluation/eval_stats/eval_stats_classification.py
opcode81: It could make sense if the user has his own defaults and only in same cases needs additional metrics. In that case, the user's default array would be altered.

MischaPanch: I removed the addition

src/sensai/evaluation/eval_stats/eval_stats_base.py
opcode81: I guess I should read your commit messages (was in a hurry).
Anyway, I agree that it feels odd to instantiate a new object only to retrieve the name.
One solution would be for all metrics without arguments to be singletons. But that feels like a bit of overkill.
How about a hybrid solution like this (best of both worlds):

    class Foo:
        metricName: str

        def __init__(self, name=None):
            """
            :param name: the name of the metric; if None use the class' metricName attribute
            """
            self.name = name if name is not None else self.__class__.metricName


    class Bar(Foo):
        metricName = "BAR"


    class Baz(Foo):
        metricName = "BAZ"

        def __init__(self, param):
            super().__init__(name=f"BAZ{param}")


    class Wrong(Foo):
        pass  # will raise AttributeError

Unlike what you state, please note that your current solution is not fully backward compatible, since you dropped the constructor of Metric. A user who created his own Metric subclass and thus calls the super constructor gets an exception.

opcode81: Since I used __class__ to refer to the class attribute above, it could even also be called `name`.

opcode81: > I guess I should read your commit messages 

Is there even a way to view all changes and the corresponding commit messages at the same time (like IntelliJ's annotate)? I don't see it anywhere.
GitHub doesn't make it easy for reviewers...
One could review on a per-commit basis, but that's not the best idea in many cases.

src/sensai/util/geometry.py
```diff
@@ -89,3 +91,64 @@ def add_edge(edge_index_pair, edge_vertex_pair):
     remaining_edges = MultiLineString(edge_vertex_pairs)
 
     return unary_union(list(polygonize(remaining_edges)))
+
+
+class SpanningTree:
```
opcode81, line 96: Since a spanning tree is a concept from graph theory (not geometry), users might not go looking for something like this here.

MischaPanch: Yeah, I thought the same. In my original repo I had it in a module called trees. When migrating, I didn't want sensAI to become too much of a graph/geometry oriented library, so I hesitated to create a new module. Do you have a suggestion for where to put it?

src/sensai/evaluation/eval_stats/eval_stats_base.py
MischaPanch: Yeah, this is really annoying. I don't know of a way for doing this in Github. That's why I mentioned this important commit in the PR's description - if you hover over the link with the mouse you'll see the message

MischaPanch: I think the latest change in c1fdc6 reaches the same goal as you proposed, right?

src/sensai/evaluation/eval_stats/eval_stats_classification.py
MischaPanch: Will be resolved in the other discussion

src/sensai/evaluation/eval_stats/eval_stats_base.py
opcode81: Basically yes. The only difference is that your variant will not crash upon instantiation if neither name is provided but instead will crash later on - when the name is first accessed.

src/sensai/util/geometry.py
opcode81: Since graph theory is essential to many AI algorithms, a package dealing with graph-related stuff would not be out of place.

opcode81: One thing to consider: It could look odd to someone reading the code, as it's not a common pattern to only conditionally set an attribute. (Also: no docstring explaining it.) That's why I opted for the variant that makes the fallback to the class attribute explicit.

src/sensai/evaluation/eval_stats/eval_stats_base.py
opcode81: One thing to consider: It could look odd to someone reading the code, as it's not a common pattern to only conditionally set an attribute. (Also: no docstring explaining it.) That's why I opted for the variant that makes the fallback to the class attribute explicit.

src/sensai/util/geometry.py
MischaPanch: cool, I'll move it to util.graph

src/sensai/evaluation/eval_stats/eval_stats_base.py
MischaPanch: Ah, I see, I'll change it to your version. It's also better that it fails on init, this is easier to discover in tests. I hadn't thought about it

PR #23, opened 2020-11-28
Improvements in torch modules: NNOptimiser without validation, improved naming/documentation

opcode81:
* Support use of NNOptimiser without applying validation
    (demonstrated in a new test)
  * Improved naming of methods and classes, particularly in loss evaluators
  * Improved documentation, clarifying the purpose of classes and methods
  * Added new class TorchModelFromModuleFactory for convenience

### DIFF
--- a/src/sensai/torch/torch_data.py
+++ b/src/sensai/torch/torch_data.py
@@ -238,10 +238,26 @@ class ClassificationVectorDataUtil(VectorDataUtil):
 class TorchDataSet:
     @abstractmethod
     def iterBatches(self, batchSize: int, shuffle: bool = False, inputOnly=False) -> Generator[Union[Tuple[torch.Tensor, torch.Tensor], torch.Tensor], None, None]:
+        """
+        Provides an iterator over batches from the data set.
+
+        :param batchSize: the maximum size of each batch
+        :param shuffle: whether to shuffle the data set
+        :param inputOnly: whether to provide only inputs (rather than inputs and corresponding outputs).
+            If true, provide a single tensor, which is to serve as a model input.
+            If false, provide a pair of tensors (i, o) with inputs and corresponding outputs.
+            Some data sets may only be able to provide inputs, in which case inputOnly=False should lead to an
+            exception.
+        """
         pass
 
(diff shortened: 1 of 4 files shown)


opcode81:
This PR addresses issue #15

src/sensai/torch/torch_data.py
```diff
@@ -238,10 +238,26 @@ def _inputOutputPairs(self, indices):
 class TorchDataSet:
     @abstractmethod
     def iterBatches(self, batchSize: int, shuffle: bool = False, inputOnly=False) -> Generator[Union[Tuple[torch.Tensor, torch.Tensor], torch.Tensor], None, None]:
```
MischaPanch, line 240: Since I will probably write some data loaders for images anyway, I wonder whether this can be geneIralized a bit and used in the rest of sensai. For example, we might have a base DataSet that generates batches of numpy arrays and convert them to tensors either in a child class or directly in the methods consuming the dataset

MischaPanch: I you agree, we can follow up in a separate issue

src/sensai/torch/torch_data.py
```diff
@@ -261,6 +277,13 @@ def __init__(self, inputTensorScaler: Optional[TensorScaler] = None, outputTenso
 
     @abstractmethod
     def provideSplit(self, fractionalSizeOfFirstSet: float) -> Tuple[TorchDataSet, TorchDataSet]:
```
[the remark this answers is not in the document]
opcode81: "fraction" would be unclear (what is a fraction of what?)

src/sensai/torch/torch_data.py
opcode81: Yes, we can add such an issue, it would make sense. But how about batches of pandas.DataFrames?

src/sensai/torch/torch_data.py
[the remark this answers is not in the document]
MischaPanch: For me the parameter docu would be enough, but I don't really care

MischaPanch (approved):

PR #27, opened 2020-11-30
Improving build, adding notebooks and minor extensions

MischaPanch:
This PR does a bunch of things at once, see the linked issues.

It turned out that in my notebooks I was relying on changes that I had not merged into sensAI yet. I had to add these changes here to make everything work... Sorry for that, I wanted to keep it purely with build-stuff.


src/sensai/data_transformation.py
```diff
@@ -331,7 +331,8 @@ def __init__(self, rules: Sequence[Rule], defaultTransformerFactory=None, requir
         """
         :param rules: the set of rules to apply
         :param defaultTransformerFactory: a factory for the creation of transformer instances (from sklearn.preprocessing, e.g. StandardScaler)
-            that shall be used to create a transformer for all rules that don't specify a particular transformer
+            that shall be used to create a transformer for all rules that don't specify a particular transformer.
+            The default transformer will only be used if requireAllHandled is False.
```
opcode81, line 335: This newly added statement is incorrect.

src/sensai/evaluation/crossval.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, data: InputOutputData, folds: int = 5, randomSeed=42, returnT
             testIndices = permutedIndices[testStartIdx:testEndIdx]
             trainIndices = np.concatenate((permutedIndices[:testStartIdx], permutedIndices[testEndIdx:]))
             self.modelEvaluators.append(self._createModelEvaluator(data.filterIndices(trainIndices), data.filterIndices(testIndices)))
+        super().__init__()
```
opcode81, line 89: Calling the super constructor should be the first (not the last) thing we do whenever possible.

src/sensai/evaluation/evaluator.py
```diff
@@ -20,10 +21,28 @@
...
+    def unsetTrackedExperiment(self):
+        self.trackedExperiment = None
+
+    def computeAndTrack(self, model, returnResult=True, **kwargs) -> Optional[Dict[str, float]]:
+        if self.trackedExperiment is None:
+            raise Exception("Cannot track results: trackedExperiment is not set.")
+        valuesDict = self.computeMetrics(model, **kwargs)
+        valuesDict["str(model)"] = str(model)
+        self.trackedExperiment.trackValues(valuesDict)
+        if returnResult:
```
opcode81, line 43: Why do we need this parameter?

tests/conftest.py
```diff
@@ -1,48 +1,48 @@
...
-
-@pytest.fixture(scope="session")
-def irisDataSet():
-    return IrisDataSet()
-
-
-@pytest.fixture(scope="session")
-def irisClassificationTestCase(irisDataSet):
-    return ClassificationTestCase(irisDataSet.getInputOutputData())
+import logging
```
opcode81, line 1: What happened to this file?

src/sensai/evaluation/evaluator.py
```diff
@@ -20,10 +21,28 @@
...
 
 class MetricsDictProvider(ABC):
+    def __init__(self):
+        self.trackedExperiment: Optional[TrackedExperiment] = None
+
     @abstractmethod
     def computeMetrics(self, model, **kwargs) -> Dict[str, float]:
         pass
 
+    def setTrackedExperiment(self, trackedExperiment: TrackedExperiment):
```
opcode81, line 31: Maybe add a docstring explaining that tracking will take place only if one explicitly calls computeAndTrack rather than computeMetrics, which is perhaps unexpected. (Why aren't we wrapping computeMetrics instead?)

opcode81 (changes requested):
Can't really comment on the build stuff. Some minor remarks on the code changes (at least one needs fixing).

src/sensai/data_transformation.py
MischaPanch: Hm, why? if requireAllHandled is true, an exception is raised. The default transformer is not applied then

src/sensai/evaluation/evaluator.py
MischaPanch: if you decide to track, you might not need to return the result. This could be called in a loop, populating some remote dashboard

tests/conftest.py
MischaPanch: I don't remember touching it, will look into that

src/sensai/evaluation/evaluator.py
MischaPanch: I did not want to make too large changes when I included it. Previously experiments were only set in local search loops. I wanted that mechanism to continue being valid, so I did not simply wrap computeMetrics. I guess I could but then the code in hyperopt becomes more convoluted, I think. I will have another look though, these changes were made months ago

src/sensai/data_transformation.py
opcode81: No. You misread the code.
Every rule that has a match but has no transformer of its own will use the default transformer.
requireAllHandled is an unrelated flag which will throw an exception if there is at least one column for which no rule was matched.

src/sensai/evaluation/evaluator.py
opcode81: Returning a value does not cost anything: at the machine level, the address of the return value (which already exists) is merely pushed onto the stack (rather than None). Omitting it is, therefore, not useful.

src/sensai/evaluation/evaluator.py
opcode81: I meant that the current method `computeMetrics` could become `_computeMetrics` and the new `computeMetrics` could call it and track the results if there is a tracked experiment. It would be completely transparent to `hyperopt`.

src/sensai/data_transformation.py
MischaPanch: Ahh, that makes a lot of sense. Ok, I will edit the docstring accordingly

src/sensai/evaluation/evaluator.py
MischaPanch: It wouldn't, since computeMetrics is called in hyperopt which potentially has its own experiment. This way the tracking might happen twice to multiple experiments or can have any kind of strange effects if the same experiment is used in evaluator and hyperopt (I believe for trains it might break). Am I missing something?

src/sensai/evaluation/evaluator.py
MischaPanch: You are right, I will remove the parameter

src/sensai/evaluation/evaluator.py
opcode81: It's transparent, because it would do what we want without any code changes.
If a user passed the same experiment to both objects, he clearly wants to track stuff twice, doesn't he?

MischaPanch: Hm, I haven't thought about it like this. I guess we can delegate this to the user, maybe logging a warning or even raising an exception when there is an experiment in both objects.

If we don't raise errors, it might lead to unexpected behaviour, at least with trains. There tracking to the same experiment twice actually overwrites the result. Passing the same experiment to both objects would completely break the tracking and passing two different ones would be nonsensical. I tend to agree with your proposal but let me try it out to see if I'm forgetting something.

MischaPanch: Ah, I remember: for trains we have to track the result of the full grid search (due to overwriting) and not of a single call of computeMetrics. I wanted to permit setting the tracked experiment in the evaluator and then running the grid search with it, without setting the experiment in the grid search. Then I cannot proceed as you suggested. Do you see this as a valid use case? To me it feels more natural to bind the experiment to the evaluator instead of the hyperopt objects

opcode81: "I wanted to permit setting the tracked experiment in the evaluator and then running the grid search with it, without setting the experiment in the grid search." - that seems contrived to me. The user should set the experiment precisely where we wants stuff to be tracked.

tests/conftest.py
MischaPanch: This is just a github bug. Nothing happened to it, I just renamed the directory test to tests (that's the pytest convention). The history is ok and the blame works as expected but github shows the whole thing as green...

src/sensai/evaluation/evaluator.py
MischaPanch: Done

src/sensai/data_transformation.py
MischaPanch: Done

src/sensai/evaluation/evaluator.py
MischaPanch: Done

src/sensai/evaluation/evaluator.py
MischaPanch: Done

src/sensai/evaluation/crossval.py
MischaPanch: I have a proposal of using a mixing for setting experiments, it's in a separate commit that can be dropped if you don't like the mechanism. It makes the code simpler, no need for inits and duplications. It also works as one would expect, since assigning a class attribute from an instance actually changes the class attribute to the instance. Only mutation of class attributes gives strange behaviour but this is not possible since None is unmutable.

The downside is, that several classes now have a class attribute `trackedExperiment` which evaluates to None. I would say it is only an aesthetical issue and I am personally not bothered by it. What do you think? Dropping the last commit is easy

MischaPanch: If you don't like the public class attribute, I could also make it private and add a public property getter in the mixin

opcode81: The implementation is somewhat "dangerous", as it does not follow the conventions of setters. A caller will assume that calling `setTrackedExperiment` on one instance has no impact on any other instance. This is not the case with the current implementation, as we are sharing static state here. In particular, it would be rather unexpected for a user to see that class A's executions are being tracked only because an instance of unrelated class B had an experiment set (as a result of class A also "being" a TrackingMixin). Furthermore, it seems limited that we can only ever have one tracked experiment at any given time; it does not seem far-fetched to want to simultaneously track several parallel experiments.

Also: I am curious what will happen in a multi-processing context here: Will a static attribute of the object being serialised carry over to the new process? I really don't know what will happen. Unless multiprocessing specifically considers this, it won't, as serialising the actual parameters of the functions being "multi-processed" will not consider static attributes.

MischaPanch: Nice, the best of both worlds! I'll change it accordingly, although it will still remain somehow hacky, I guess

MischaPanch: Done. Although it does not really change anything, right? If the instance attribute is not set, if will return the class attribute which was None

opcode81: It improves the code quality and requires a less "sharp eye": If you see class attribute `_foo` and you see `self._foo = bar`, you assume that the class attribute is being set (even though it's not).
But functionally, yes, it's the same.

opcode81: Your update didn't remove the class attribute. It's not required anymore.

MischaPanch: Done. Even before reading your second comment ^^.  Pycharm complains now but we have to let it complain

opcode81: If you want to avoid the warnings, always use `setattr` and `getattr`, ideally with a class-level attribute for the name.

MischaPanch: Done. I wouldn't necessarily say it's much cleaner but the warnings are gone

opcode81: Importantly, it's significantly cleaner than the shadowed attribute, which is the main thing to improve upon.

MischaPanch: Absolutely! I meant that `setattr` was not much cleaner than doing `self._trackedExperiment = ...`. Removing the class-attribute was definitely worth it!

opcode81 (approved):

PR #33, opened 2020-12-06
Unsupervised and rule-based models

MischaPanch:
First step towards supporting those use cases. The idea is that such models also
require feature engineering, transformations and so on. They should also
be compatible with evaluators and cross validators. Evaluation interfaces
will be adjusted in a following commit.

The changes are essentially backwards compatible as only abstract classes are adjusted
and some args were given a default value. Only custom implementations of PredictorModel
will truly be affected but in a rather non-intrusive way.

### DIFF
--- a/src/sensai/featuregen.py
+++ b/src/sensai/featuregen.py
@@ -104,7 +104,7 @@ class FeatureGenerator(ABC):
         return self._generatedColumnNames
 
     @abstractmethod
-    def fit(self, X: pd.DataFrame, Y: pd.DataFrame, ctx=None):
+    def fit(self, X: pd.DataFrame, Y: pd.DataFrame = None, ctx=None):
         """
         Fits the feature generator based on the given data
 
@@ -172,7 +172,7 @@ class FeatureGenerator(ABC):
         """
         pass
 
-    def fitGenerate(self, X: pd.DataFrame, Y: pd.DataFrame, ctx=None) -> pd.DataFrame:
+    def fitGenerate(self, X: pd.DataFrame, Y: pd.DataFrame = None, ctx=None) -> pd.DataFrame:
         """
         Fits the feature generator and subsequently generates features for the data points in the given data frame
 
@@ -191,7 +191,7 @@ class RuleBasedFeatureGenerator(FeatureGenerator, ABC):
     """
     A feature generator which does not require fitting
     """
-    def fit(self, X: pd.DataFrame, Y: pd.DataFrame, ctx=None):
+    def fit(self, X, Y=None, ctx=None):
         pass
 
 
@@ -229,12 +229,12 @@ class MultiFeatureGenerator(FeatureGenerator):
             return fg.generate(inputDF, ctx=ctx)
         return self._generateFromMultiple(generateFeatures, inputDF.index)
 
-    def fitGenerate(self, X: pd.DataFrame, Y: pd.DataFrame, ctx=None) -> pd.DataFrame:
+    def fitGenerate(self, X: pd.DataFrame, Y: pd.DataFrame = None, ctx=None) -> pd.DataFrame:
         def generateFeatures(fg: FeatureGenerator):
             return fg.fitGenerate(X, Y, ctx)
         return self._generateFromMultiple(generateFeatures, X.index)
 
-    def fit(self, X: pd.DataFrame, Y: pd.DataFrame, ctx=None):
+    def fit(self, X: pd.DataFrame, Y: pd.DataFrame = None, ctx=None):
         for fg in self.featureGenerators:
             fg.fit(X, Y)
 
@@ -411,10 +411,10 @@ class ChainedFeatureGenerator(FeatureGenerator):
             df = featureGen.generate(df, ctx)
         return df
 
-    def fit(self, X: pd.DataFrame, Y: pd.DataFrame, ctx=None):
+    def fit(self, X: pd.DataFrame, Y: pd.DataFrame = None, ctx=None):
         self.fitGenerate(X, Y, ctx)
 
-    def fitGenerate(self, X: pd.DataFrame, Y: pd.DataFrame, ctx=None) -> pd.DataFrame:
+    def fitGenerate(self, X: pd.DataFrame, Y: pd.DataFrame = None, ctx=None) -> pd.DataFrame:
         for fg in self.featureGenerators:
             X = fg.fitGenerate(X, Y, ctx)
         return X
@@ -466,7 +466,7 @@ class FeatureGeneratorTargetDistribution(FeatureGenerator):
         # This will hold the mapping: column -> featureValue -> targetValue -> targetValueEmpiricalProbability
         self._discreteTargetDistributionsByColumn: Dict[str, Dict[Any, Dict[Any, float]]] = None
 
-    def fit(self, X: pd.DataFrame, Y: pd.DataFrame, ctx=None):
+    def fit(self, X: pd.DataFrame, Y: pd.DataFrame = None, ctx=None):
         """
         This will persist the empirical target probability distributions for all unique values in the specified columns
         """
@@ -657,7 +657,7 @@ class FeatureGeneratorFromVectorModel(FeatureGenerator):
         self.useTargetFeatureGeneratorForTraining = useTargetFeatureGeneratorForTraining
         self.vectorModel = vectorModel
 
-    def fit(self, X: pd.DataFrame, Y: pd.DataFrame, ctx=None):
+    def fit(self, X: pd.DataFrame, Y: pd.DataFrame = None, ctx=None):
         targetDF = self.targetFeatureGenerator.fitGenerate(X, Y)
         if self.inputFeatureGenerator:
             X = self.inputFeatureGenerator.fitGenerate(X, Y)
--- a/src/sensai/vector_model.py
+++ b/src/sensai/vector_model.py
@@ -15,10 +15,35 @@ log = logging.getLogger(__name__)
 T = TypeVar('T')
 
 
-class PredictorModel(ABC):
+class _IdentityFG(FeatureGenerator):
+    def _generate(self, df: pd.DataFrame, ctx=None) -> pd.DataFrame:
+        return df
+
+    def fit(self, X, Y=None, ctx=None):
+        pass
+
+
+class _IdentityDFT(InvertibleDataFrameTransformer):
+    def applyInverse(self, df: pd.DataFrame) -> pd.DataFrame:
+        return df
+
+    def fit(self, df):
+        pass
+
+    def apply(self, df: pd.DataFrame) -> pd.DataFrame:
+        return df
+
+
+class PredictorModel(PickleLoadSaveMixin, ABC):
     """
-    Base class for models that map vectors to predictions
+    Base class for models that map data frames to predictions
     """
+    def __init__(self):
+        self._featureGenerator: FeatureGenerator = _IdentityFG()
+        self._inputTransformerChain = DataFrameTransformerChain(())
+        self._outputTransformerChain = DataFrameTransformerChain(())
+        self._name = None
+
     @abstractmethod
     def predict(self, x: pd.DataFrame) -> pd.DataFrame:
         pass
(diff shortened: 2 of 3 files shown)


MischaPanch:
Tests run through, see corresponding action. It is not linked here since the PR is to develop and we have to merge the other PR first for the new builds to be triggered

src/sensai/vector_model.py
```diff
@@ -164,41 +251,50 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
...
+            )
+        self._targetTransformer.fit(Y)
+        Y = self._targetTransformer.apply(Y)
+
         self._modelInputVariableNames = list(X.columns)
-        self._modelOutputVariableNames = list(Y.columns)
-        log.info(f"Training with outputs[{len(self._modelOutputVariableNames)}]={self._modelOutputVariableNames}, inputs[{len(self._modelInputVariableNames)}]=[{', '.join([n + '/' + X[n].dtype.name for n in self._modelInputVariableNames])}]")
+        output_info = ""
+        if Y is not None:
+            self._predictedVariableNames = list(Y.columns)
```
opcode81, line 281: This is broken. Please never change the order of statements without due cause. predictedVariableNames and modelOutputVariableNames are not the same thing. For an explanation, see the docstring of getModelOutputVariableNames.

src/sensai/vector_model.py
```diff
@@ -253,7 +326,11 @@ def __init__(self):
     def isRegressionModel(self) -> bool:
         return False
 
-    def _fit(self, X: pd.DataFrame, Y: pd.DataFrame):
+    def _fit(self, X: pd.DataFrame, Y: Optional[pd.DataFrame]):
+        if Y is None:
+            raise NotImplemented(
+                "Unsupervised classification models are currently not implemented"
+            )
```
opcode81, line 333: Continued statements must always use indentation. Please do not use this style of formatting.

src/sensai/vector_model.py
```diff
@@ -133,20 +127,113 @@ def withName(self, name: str):
...
+
+
+class FittableModel(PredictorModel, ABC):
     @abstractmethod
-    def isRegressionModel(self) -> bool:
+    def fit(self, X: pd.DataFrame, Y: pd.DataFrame = None):
+        pass
+
+
+class RulesBasedModel(FittableModel, ABC):
```
opcode81, line 157: Since the commonly used adjective is "rule-based", we should omit the 's'.

src/sensai/vector_model.py
```diff
@@ -133,20 +127,113 @@ def withName(self, name: str):
...
+    """
+    Base class for models where the essential logic is rule based but the feature generators
```
opcode81, line 159: suggestion: ... the essential prediction logic is based on rules coded by humans and thus does not require fitting. However, the input generation process may use mechanisms such as feature generation or data transformation, which may require fitting.

src/sensai/vector_model.py
opcode81, line 157: I don't really see the need for this class, especially since it duplicates functionality from VectorModel for no good reason - and loses some functionality along the way. For instance, the checks done in _computeInputs are omitted here. 
Why wouldn't it be OK to implement a subclass of VectorModel with an empty _fit? After all, a rule-based model _is_ a regular model which just doesn't do fitting. Either way, it would be more natural for this to be a specialisation rather than a sibling in the hierarchy.

src/sensai/vector_model.py
```diff
@@ -15,10 +15,35 @@
 T = TypeVar('T')
 
 
-class PredictorModel(ABC):
+class _IdentityFG(FeatureGenerator):
```
opcode81, line 18: What exactly is the reason for these classes? Why did you think it wasn't appropriate to use None-checks (not all of which were removed) instead?
None-checks being removed breaks backward compatibility (and loses convenience), because it would no longer be OK to pass None to the respective with* method (which can be very convenient for flexibly configured model construction). Granted, the possibility to pass None is not properly type annotated, but neither are your changes (e.g. regarding Y in this PR). It really is annoying to have to add Optional, isn't it?

opcode81 (changes requested):

src/sensai/vector_model.py
MischaPanch: Implementing empty _fit compromises efficiency as during fit we would call fitGenerate on the feature generators instead if simply fit. I don't see how I can make it a specialization of VectorModel because of that. The RuleBasedModel cannot have the input checks for the above reason

MischaPanch: Yeah, that was a typo propagated by autocompetion ^^

opcode81: I see. If efficiency is more important (which is certainly can be), I guess it's the right decision you made.

opcode81: Then again, don't we usually have time to spare when training?

src/sensai/featuregen.py
```diff
@@ -104,7 +104,7 @@ def getGeneratedColumnNames(self) -> Optional[List[str]]:
         return self._generatedColumnNames
 
     @abstractmethod
-    def fit(self, X: pd.DataFrame, Y: pd.DataFrame, ctx=None):
+    def fit(self, X: pd.DataFrame, Y: pd.DataFrame = None, ctx=None):
```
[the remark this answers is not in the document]
MischaPanch: I was thinking that the next step would be to allow for the abstract subclasses of VectorModel to also be unsupervised. The actual implementations of supervised models can forbid passing None in _fit. As for feature generators, I do think it is very natural to allow passing None for Y. Even when they are fit on input data it does not necessarily need to involve the target.

I could make an UnsupervisedModel, of course, but it would involve duplicating some code or overwriting the fit (or some more serious refactoring). Also, if we were to follow my plan above, it would make the hierarchy more complicated, as we would need to duplicate all child classes. We then have an UnsupervisedClassificationVectorModel, UnsupervisedRegressionVectorModel, and so on differing only by whether Y is passed or not. 

Maybe we can have a short sync on the phone about the best way to include unsupervised models and generators?

src/sensai/vector_model.py
MischaPanch: In my use case it is. For the RuleBasedModel the rules are very simple (a sum) but feature generator is costly (processing images). In fact, most feature generation will be rule based itself so fit is for free but fitGenerate is expensive

src/sensai/vector_model.py
MischaPanch: Yeah, I was unsure about that myself. It felt good not to have None checks everywhere in the RuleBasedModel 
(fit if not none, generate if not None and so on). We also have not-None input and output DTF transformers, so this change felt natural. I was thinking about backwards compatibility but I considered that it would be rather a nuisance. Do you perform such checks in your code at the moment?

MischaPanch: Concerning the incorrect type annotations, I need to check. Maybe you can give a more specific example? For kwargs with default None the Optional thing is inferred automatically

src/sensai/vector_model.py
MischaPanch: Nice catch, I did not want that to happen. Will go through the code again (once we settled on how to deal with unsupervised models)

src/sensai/vector_model.py
MischaPanch: So you are not a fan of black formatter 😄 . Ok, I will remove it.

The idea is to minimize the diff, so here a change in the string is on a separate line and one immediately see that nothing else has changed

src/sensai/vector_model.py
MischaPanch: Btw, I did allow passing None in the with* methods

src/sensai/vector_model.py
```diff
@@ -133,20 +127,113 @@ def withName(self, name: str):
...
+
+    def predict(self, x: pd.DataFrame) -> pd.DataFrame:
+        """
+        Performs a prediction for the given input data frame
+
+        :param x: the input data
+        :return: a DataFrame with the same index as the input
+        """
+        if not self._isFitted:
+            log.warning(
```
opcode81, line 179: We could avoid the warning in cases where there is no data transformation and no feature generation.
I am on the fence as to whether in cases where there are, this should be an exception; the logged warning could remain unnoticed.

MischaPanch:
The nuisance I meant was that my changes broke user code of the type:

```
if myVectorModel.targetTransformer is None: 
     doSomething
```

I don't think this will happen anywhere, so I was not too afraid of breaking it.

>Indeed, they were probably all kwargs, so they're fine.
>I don't understand your statements above: What would be a nuisance, and which checks am I using? The scenario I was ?>referring to is this:
>
>targetTransformer = None
>if useTargetTransformation:
>   targetTransformer = MyTransformer()
>
>model = MyModel().withTargetTransformer(targetTransformer)

src/sensai/vector_model.py
MischaPanch: I don't agree here, it will be common that a RuleBasedModel only has RuleBasedFgen and DFTs. Fitting does nothing then and the user shoudn't be forced to do that. In any case for rule based models there is (necessarily) less safety and the user has to be a bit more aware of what they are doing

src/sensai/vector_model.py
MischaPanch: There were a few None checks in VectorModel but then I needed a few more in the RuleBased one and we might need more for further abstractions. The added complexity ist arguable, I would say - for me the code in fit and predict looks significantly neater without the additional None checks. I think that this code which will be read much more often than the builder methods setting the fgen or target transformer or these two trivial classes. Also, it is in line with having _inputTransformerChain and _outputTransformerChain not being None even when they are trivial. So overall, the tradeoff of having two private classes vs None checks seemed beneficial to me

src/sensai/vector_model.py
MischaPanch: Done

src/sensai/featuregen.py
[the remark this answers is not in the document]
MischaPanch: Yeah, I haven't completely thought this through. I reverted the relaxation of the interface in the models. As discussed, I left the relaxed interface for the feature generators, where it does make sense

src/sensai/vector_model.py
MischaPanch: Done by reverting to previous version of the code

src/sensai/vector_model.py
MischaPanch: Removed this code

src/sensai/vector_model.py

src/sensai/vector_model.py
```diff
@@ -133,27 +125,119 @@ def withName(self, name: str):
...
+
+    def setName(self, name):
+        self._name = name
+
+    def getName(self):
+        if self._name is None:
+            return "unnamed-%s-%x" % (self.__class__.__name__, id(self))
+        return self._name
+
+    # TODO: why do we need the getter and the setter? We only have them for fgens, not for the input/output/target transformers
```
MischaPanch, line 142: @opcode81 What do you think of this? I would be in favor of removing the getter and setter, or do you see good use cases for them?

src/sensai/vector_model.py
```diff
@@ -164,10 +248,10 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
         """
         x = self._computeInputs(x)
         y = self._predict(x)
+        # TODO or not TODO: why do we do this? Isn't it the implementation's responsibility to get it right?
```
MischaPanch, line 251: @opcode81 What do you think? If we remove this line, we could save some ram (negligible in most cases, of course) by always redeclaring x instead of having x and y.

src/sensai/vector_model.py
opcode81: As a user, I would object to receiving warnings that do not apply. Ideally, we should not warn that there might be a problem but _know_ if there is a problem (by having the components be aware of the fitting requirement) - and throw an exception iff so.

opcode81: Maybe this is material for an improvement that adds the "knowledge" to DFTs, etc.

MischaPanch: I agree. Let's maybe address it in a separate issue? I don't want to further blow up this PR

src/sensai/vector_model.py
opcode81: What would be the benefit of removing them?
The getter was used in at least one project, and the setter, while functionally redundant, makes sense for post-construction use, where writing model.withFG(fg) would feel weird.

src/sensai/vector_model.py
opcode81: Currently, implementations don't correctly set the index, so the line is necessary. It would be a source of error in general, because you will never forget to set the contents of the DF, yet you may not realise that setting the index could even matter.
RAM could be saved only in the case where _computeInputs is not id.

opcode81: To make this "cleaner", we could consider changing the interface of `_predict` to not return a DF but a more low-level result instead and make `predict` construct the DF correctly.

src/sensai/vector_model.py
opcode81: Yes, separate issue.

src/sensai/vector_model.py
MischaPanch: It just felt strange that this exists only for fgen. If we want to allow getter and setter, why not for all other components as well?

src/sensai/vector_model.py
MischaPanch: I'll open a separate issue which we may or may not resolve some time

src/sensai/vector_model.py
opcode81: The reason is that sensAI features were added only when needed.
If you think that users could reasonably want these getters/setters, you can add them, but you should have a reasonable use case in mind. If the use case is entirely "hacky", then we shouldn't add them, as users can make do with equally hacky private access. :)

MischaPanch: That's what I suspected ^^. Ok, let's leave it as is, certainly there is no harm.

opcode81 (approved):

PR #41, opened 2021-01-13
Feature/more informative featuregens and dtftransf

### LINKED ISSUE #35: Objects such as DFTs and Fgens should know whether they actually require fitting ...
As a user, I would object to receiving warnings that do not apply. Ideally, we should not warn that there might be a problem but _know_ if there is a problem (by having the components be aware of the fitting requirement) - and throw an exception iff so.

_Originally posted by @opcode81 in https://github.com/jambit/sensAI/pull/33#discussion_r539341929_

MischaPanch:
Closes #35 

There is a bunch of breaking changes but I don't see a good way around them. They are rather on the benign side though, the model-loading regression test works (EDIT: in fact not quite, see my comment in the code below). In some implementations fit/apply would need to be changed to _fit/_apply to be compatible again.

Even though this focuses on the linked issue, it is a necessary step towards generator fitting, since typically preprocessors like one-hot-encoders, scalers and so on cannot be fitted batch-wise. I have indicated how the new features could be used for generator fitting

I have to say the None-checks for fgens and target transformers did get quite annoying here ;)


src/sensai/featuregen.py
```diff
@@ -707,4 +785,4 @@ def flattenedFeatureGenerator(fgen: FeatureGenerator, columnsToFlatten: List[str
         return ChainedFeatureGenerator(fgen, flatteningGenerator)
     else:
         return ChainedFeatureGenerator(fgen,
```
MischaPanch, line 787: Because of this change in the signature the regression test fails. I could fix it by having a more flexible signature with the _flatten method (see DFTChain) but I would prefer to go into another direction and start having more consistent interfaces even if it means breaking changes. Here the inconsistency was between ChainedFeatureGenerator, MultiFeatureGenerator and ChainedDataFrameTransformer. There were some more, which I highlighted by TODOs. 

What do you think, Dominik?

src/sensai/data_transformation.py
```diff
@@ -39,44 +100,73 @@ def applyInverse(self, df: pd.DataFrame) -> pd.DataFrame:
...
         for transformer in self.dataFrameTransformers:
-            if fit:
-                transformer.fit(df)
             df = transformer.apply(df)
         return df
 
-    def fit(self, df: pd.DataFrame):
-        for transformer in self.dataFrameTransformers:
-            transformer.fit(df)
+    def _fit(self, df: pd.DataFrame):
```
MischaPanch, line 139: This changes the logic but I believe the previous logic was incorrect, right? Apply happens in a chain, so fit should do so as well

src/sensai/vector_model.py
```diff
@@ -247,24 +259,44 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
...
+    #  Pre- and postprocessors can typically not be fit on batches which is one of the reasons I added the flag above.
+    def fitDataSet(self, dataSet):
+        """
+        Fits the models on a DataSet which does not need to be loaded to ram. Any pre- and postprocessors forming
+        part of the model are expected to be fit prior to calling this method or to not require fitting.
+        :param dataSet:
+        :return:
+        """
+        for batch in dataSet:
+            self.fit(batch.X, batch.Y, fitPreprocessors=False, fitTargetTransformer=False)
```
opcode81, line 299: This won't work. We can't just call fit several times. The learning algorithm will usually want to do the iteration over batches.

src/sensai/data_transformation.py
```diff
@@ -20,15 +21,75 @@ class DataFrameTransformer(ABC):
     (possibly applying the transformation to the original data frame - in-place transformation).
     A data frame transformer may require being fitted using training data.
     """
+    def __init__(self):
```
opcode81, line 24: Adding these properties breaks compatibility with persisted instances.
`__setstate__` should be implemented to add these properties and `__getstate__` should be implemented to not store these features, as they represent unnecessary transient state.

src/sensai/data_transformation.py
```diff
@@ -20,15 +21,75 @@ class DataFrameTransformer(ABC):
...
+        if inputColumns != outputColumns:
+            self._changesColumns = True
+            self.__removedColumns = inputColumns.difference(outputColumns)
+            self.__addedColumns = outputColumns.difference(inputColumns)
+        return df
+
+    def getChangeInColumnNames(self):
+        """
+        Returns a dict describing the change in column names that was created in the most recent application of
+        the data frame transformers. If no changes in column names were created or apply was never called, returns None.
```
opcode81, line 66: It doesn't seem appropriate to return None in both of these cases. If the DFT has never been applied, an exception seems more appropriate.

src/sensai/data_transformation.py
opcode81: Yes, definitely!

src/sensai/data_transformation.py
```diff
@@ -224,17 +337,26 @@ def apply(self, df: pd.DataFrame):
...
+    def summary(self):
+        summary = super().summary()
+        summary["columns"] = self.columns
+        summary["inplace"] = self.inplace
+        summary["handleUnknown"] = self.handleUnknown
+        summary["inferCategories"] = self.inferCategories
+        return summary
 
+
+# TODO: we have a feature generator doing the same thing. Can we remove this and its child below?
```
opcode81, line 349: No. Having a feature generator with the same functionality does not make these transformers obsolete. DFTs are used for a different purpose, and in particular, are chainable, whereas FeatureGenerators are not.
Also: compatibility is more important than reduced redundancy.

src/sensai/data_transformation.py
```diff
@@ -395,67 +518,112 @@ def apply(self, df: pd.DataFrame) -> pd.DataFrame:
...
             df[series.name] = series
         return df
 
+    def summary(self):
+        summary = super().summary()
+        summary["inplace"] = self.inplace
+        return summary
 
+
+# TODO: shouldn't this and the aggregation below rather be feature extractors?
```
opcode81, line 548: No. For the same reason as above.

opcode81 (changes requested):
Many nice improvements here, but some things still need attention.

src/sensai/featuregen.py
```diff
@@ -194,16 +216,21 @@ class RuleBasedFeatureGenerator(FeatureGenerator, ABC):
...
 class MultiFeatureGenerator(FeatureGenerator):
     """
     Wrapper for multiple feature generators. Calling generate here applies all given feature generators independently and
     returns the concatenation of their outputs
     """
-    def __init__(self, featureGenerators: Sequence[FeatureGenerator]):
-        """
-        :param featureGenerators:
-        """
+    def __init__(self, *featureGenerators: FeatureGenerator):
```
opcode81, line 231: This is not an acceptable breaking change.

opcode81 (changes requested):

src/sensai/featuregen.py
opcode81: Definitely the flatten variant. Otherwise all persisted instances will be broken.

opcode81: However, I don't see how this change really creates more consistency in any way. Just because we happen to use a *list interface in one place doesn't mean we need to have it everywhere. There are many places where lists are expected, and it is often appropriate to make that list explicit (especially if the list isn't the only argument).
At the end of the day, you will immediately see what is required in your IDE anyway, so it's "difficult" to make a mistake. While I do think it's rather cumbersome, the flatten variant is OK for me, but I really don't think that adding square brackets is a big deal - and it's not really worse than having to add a star. What is more usable boils down to what you are usually likely to have. In the case of MultiFeatureGenerator, it was common to have the list, which is why it has the interface it has.

src/sensai/data_transformation.py
```diff
@@ -39,44 +100,73 @@ def applyInverse(self, df: pd.DataFrame) -> pd.DataFrame:
...
+                result.append(x)
+            else:
+                result.extend(x)
+        return result
 
-    def apply(self, df: pd.DataFrame, fit=False) -> pd.DataFrame:
-        """
-        Applies the chain of transformers to the given DataFrame, optionally fitting each transformer before applying it.
-        Each transformer in the chain receives the transformed output of its predecessor.
+    # TODO: do we really need the more flexible interface? I propose a breaking change to *dataFrameTransformers: DataFrameTransformer
```
opcode81, line 128: Why can't we just keep the old interface? What was the reason for the change in the first place?

opcode81 (changes requested):

src/sensai/featuregen.py
MischaPanch: Thanks for the lightning fast review! I will resort to the flatten variant to permit both interfaces then

src/sensai/featuregen.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, categoricalFeatureNames: Optional[Union[Sequence[str], str]]
             raise ValueError(f"normalisationRules should be empty when a normalisationRuleTemplate is provided")
 
         self._generatedColumnNames = None
+        self.categoricalFeatureNames = categoricalFeatureNames
```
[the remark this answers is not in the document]
MischaPanch: sure, I just wanted it for the summary. 

Short story on why I did it: I don't always make things private assuming "the user will know not to change it". I remember that for some base classes in sklearn all parameters should be saved as public fields of the same name (forgot what for now), so the habit sticked

src/sensai/featuregen.py
```diff
@@ -441,7 +506,7 @@ def __init__(self, columns: Union[str, Sequence[str]], targetColumn: str,
         """
         :param columns: the categorical columns for which to generate distribution features
         :param targetColumn: the column the distributions over which will make up the features.
-            If targetColumnBins is not None, this column will be discretised before computing the conditional distributions
+            If targetColumnBins is not None, this column will be discretized before computing the conditional distributions
```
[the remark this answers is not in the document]
MischaPanch: Splendid, will do :)

src/sensai/featuregen.py
```diff
@@ -194,16 +216,21 @@ class RuleBasedFeatureGenerator(FeatureGenerator, ABC):
...
     """
     Wrapper for multiple feature generators. Calling generate here applies all given feature generators independently and
     returns the concatenation of their outputs
     """
-    def __init__(self, featureGenerators: Sequence[FeatureGenerator]):
-        """
-        :param featureGenerators:
-        """
+    def __init__(self, *featureGenerators: FeatureGenerator):
+        if len(featureGenerators) == 0:
```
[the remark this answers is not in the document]
MischaPanch: this was for consistency again - the chain one can't be empty (while the DFTchain can be, causing more potential confusion)

src/sensai/featuregen.py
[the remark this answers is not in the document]
opcode81: It's not only about knowing not to change it but also about assuming what it contains. Even reading it could lead to problems. A user might think it's always a list of column names. Since you can't provide a docstring for an attribute (what a broken language!), users can't even safely read the property.

src/sensai/vector_model.py
```diff
@@ -112,6 +103,10 @@ def withName(self, name: str):
         self.setName(name)
         return self
 
+    # TODO: why do we need this? As a user I usually would not want to look for specific classes.
```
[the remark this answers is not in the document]
MischaPanch: Right, I understand that. But assuming we "go more forward" with sensAI and sort of publicize it, coherence and removing of redundancy become more important than compatibility with earlier, non-publicly released versions.

Now I get that we are in different situations - I am only using sensai in one project, which I keep updating constantly, whereas you do so in many longer running ones. How about a compromise: we keep the todo for the while being and before we release the first really stable API, we go over them again?

src/sensai/vector_model.py
```diff
@@ -162,12 +165,9 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
         :param x: the input data
         :return: a DataFrame with the same index as the input
         """
-        if not self._isFitted:
```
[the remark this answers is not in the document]
MischaPanch: The feature generator raises the exception itself when appropriate, one on the improvements of this PR

src/sensai/vector_model.py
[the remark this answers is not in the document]
opcode81: This is not harmful in any way, not even harmful in the sense that it could be perceived as inconsistent or incoherent.

src/sensai/data_transformation.py
MischaPanch: Currently I use it for the summary, which should never raise exceptions. This is also in line with a similar method in FeatureGenerator

```
    def getGeneratedColumnNames(self) -> Optional[List[str]]:
        """
        :return: Column names of the data frame generated by the most recent call of the feature generators 'generate' method.
            Returns None if generate was never called.
        """
        return self._generatedColumnNames
```
This also returns None if generate was never called

src/sensai/data_transformation.py
MischaPanch: Same question as above, how about we leave the TODO?

While I agree that it is a bit contrived, ChainedFeatureGenerator is a pretty useful concept in my opinion. At least I tend to use it quite a lot

src/sensai/featuregen.py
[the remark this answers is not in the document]
opcode81: I don't agree with your notion of consistency in general.
DFTs and FGs are essentially unrelated. There is no need for consistency across unrelated modules.
What we need is, for example, consistency of the implementation with the method documentation.

src/sensai/data_transformation.py
MischaPanch: Just consistency/simplicity (I am always thinking about the "released stable API"). Since this is addressed in other comments as well, I am resolving this conversation.

src/sensai/featuregen.py
MischaPanch: Will be resolved by using the flattening mechanism everywhere for now

src/sensai/data_transformation.py
opcode81: Then you should either remove it from the summary or find another way of differentiating these cases. It's a relevant distinction in my opinion.

opcode81: It is not in line with the method you mention, because the method returns None only if it has never been called (and not in the case where nothing was generated in the most recent call).

src/sensai/data_transformation.py
opcode81: Consistency with what?
Simplicity: no, the old interface is much simpler.

src/sensai/featuregen.py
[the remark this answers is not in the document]
MischaPanch: The inconsistency was between feature generators: ChainedFeaturegen could not be empty and Multi could before this change. They should be similar in that respect, right?

src/sensai/vector_model.py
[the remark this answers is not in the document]
MischaPanch: The practical reason for the todo was that I wanted to write a getter for the input transformer chain which should be called `getInputTransformer_S_`, only differing by the last letter. Having both methods was aesthetically unpleasing for me ^^

src/sensai/featuregen.py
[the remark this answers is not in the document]
opcode81: No, ChainedFG and MultiFG are two distinct classes which can use two distinct construction mechanisms.
The only thing that needs to be consistent is the implementation of the construction mechanisms and its docstring.

src/sensai/data_transformation.py
MischaPanch: Right, I overlooked that. I will return an empty set in case no difference was generated

src/sensai/featuregen.py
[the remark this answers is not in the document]
MischaPanch: Ok, I'll revert it.

src/sensai/vector_model.py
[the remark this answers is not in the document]
opcode81: A name like `getInputTransformers` would suggest that the result is a list of input transformers or an instance of a class InputTransformers. If you want to return the chain, a better name would be `getInputTransformerChain`.
That being said, I don't share your impression of this being aesthetically unpleasing. The functions even have different signatures (one arg vs. no arg). In languages like Java or C++, you could even have several functions with the exact same name that differ only by their input signature.

src/sensai/featuregen.py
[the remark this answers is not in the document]
opcode81: What surprises me here is that you didn't find it aesthetically unpleasing or inconsistent to add a new property which is now the only property that isn't private :)
scnr

opcode81 (changes requested):

src/sensai/vector_model.py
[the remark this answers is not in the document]
MischaPanch: Yeah, I miss method overloading a lot in python... In java I would expect such things to happen, of course. I will go fot the chain variant then

MischaPanch:
> What surprises me here is that you didn't find it aesthetically unpleasing or inconsistent to add a new property which is now 
> the only property that isn't private :)
> scnr

True that, there is a hypocrite somewhere within me ^^

src/sensai/data_transformation.py
opcode81: We shouldn't leave the TODO, because the point is that DFTs are useful in their own right (first point I made above). Someone might use DFTs in a completely different context (where FGs are not a thing) and then might need this functionality.

src/sensai/featuregen.py
opcode81: Perhaps add the flatten function to a util module?

src/sensai/vector_model.py
MischaPanch: You are right, I didn't think this through. I will just remove it for now, the work will happen in another PR. Do you have an idea for how to realize it without changing the model implementations? I don't see a straightforward way now, but it might be because its late and I am tired. The only thing coming to mind is to do some kind of aggregation of batches before we pass it to the model but it partially defeats the purpose of having a generator...

MischaPanch: Feel free to write me on wire, the discussion is out of scope for this PR

src/sensai/featuregen.py
MischaPanch: Definitely!

src/sensai/data_transformation.py
MischaPanch: True that, removing it

src/sensai/data_transformation.py
MischaPanch: Removed

src/sensai/vector_model.py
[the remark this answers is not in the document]
MischaPanch: Done. I added the getters

src/sensai/data_transformation.py
MischaPanch: Done

src/sensai/data_transformation.py
MischaPanch: Done

src/sensai/data_transformation.py
MischaPanch: Consistency with VectorModel here. The with... methods used this more relaxed interface while the chain had a different one. Anyway, I will use the to-be-created util.flatten method here (tomorrow).

opcode81 (changes requested):

src/sensai/data_transformation.py
MischaPanch: Done, I think (please have a critical look). I did the same for FeatureGenerator. I don't think the df changes history should be transient, so with the recent change there are not transient attributes.

src/sensai/data_transformation.py
```diff
@@ -54,18 +54,17 @@ def apply(self, df: pd.DataFrame) -> pd.DataFrame:
         df = self._apply(df)
         outputColumns = set(df.columns)
 
-        if inputColumns != outputColumns:
-            self._changesColumns = True
-            self.__removedColumns = inputColumns.difference(outputColumns)
-            self.__addedColumns = outputColumns.difference(inputColumns)
+        self.__removedColumns = inputColumns.difference(outputColumns)
```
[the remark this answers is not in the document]
MischaPanch: I liked the idea so much that I went in full over engineering mode and wrote a class for keeping track of changes as well as a decorator using it. It is in a the new module util.metadata. My hope is that this addition can be useful in other situations, e.g. in a chain of transformations where we wanna keep track of what is happening

src/sensai/data_transformation.py
MischaPanch: I understand your point and indeed it is uniformity, not consistency that I am after. Sorry for confusing the two, they are different notions indeed. It is true that uniformity is not necessarily an improvement and this particular change is mostly about my personal preference (I wanted to be able to use the same interface in Chained and Multi fgens). Since we agreed on the flattened solution and we have other open threads addressing it, I propose to close this thread

src/sensai/featuregen.py
MischaPanch: Done

src/sensai/data_transformation.py
MischaPanch: My initial version was wrong but with the last commit it should work (the backwardscompat test runs through). I removed the double underscore since it mangles the attribute name and becomes a bit messy in setstate. Hopefully the single underscore is enough to symbolize that _isFitted should not be accessed directly unless the user knows exactly what they are doing

src/sensai/featuregen.py
MischaPanch: Done

opcode81 (changes requested):

src/sensai/data_transformation.py
```diff
@@ -45,30 +51,27 @@ def _fit(self, df: pd.DataFrame):
     def _apply(self, df: pd.DataFrame) -> pd.DataFrame:
         pass
 
+    @trackDFHistory
```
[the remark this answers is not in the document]
MischaPanch: I was almost sure you would say it and after I finished I thought similarly :D. Will remove it, was just nice to write a decorator again

src/sensai/util/metadata.py
```diff
@@ -0,0 +1,76 @@
...
+    >>> df.index = ["first", "second"]
+    >>> dfHistory.update(df)
+    >>> dfHistory.indexHistory
+    [RangeIndex(start=0, stop=2, step=1), Index(['first', 'second'], dtype='object')]
+    >>> dfHistory.columnsHistory
+    [Index(['bar'], dtype='object'), Index(['bar', 'foo'], dtype='object')]
+    """
+    def __init__(self, df: pd.DataFrame):
+        self.numUpdates = -1  # after init this will be 0, so init does not count as update
+        self.columnsHistory = []
```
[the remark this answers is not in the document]
MischaPanch: In case we use this in a chain of transformations. Then we can see how stuff was changed in each step

src/sensai/util/metadata.py
```diff
@@ -0,0 +1,76 @@
...
+        self.indexHistory = []
```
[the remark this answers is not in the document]
MischaPanch: Yeah, I was thinking about that too. My idea was to potentially use it in the DFTs that filter out rows but I guess it´s better to write an extension for that case instead of just saving them generally

src/sensai/data_transformation.py
```diff
@@ -45,30 +51,27 @@ def _fit(self, df: pd.DataFrame):
...
-
-        inputColumns = set(df.columns)
         df = self._apply(df)
-        outputColumns = set(df.columns)
-
-        self.__removedColumns = inputColumns.difference(outputColumns)
-        self.__addedColumns = outputColumns.difference(inputColumns)
         return df
 
     def getChangeInColumnNames(self):
```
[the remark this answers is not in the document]
MischaPanch: The change in columns can be quite large, making it difficult to decipher in a string.  I also might want to use the actual list of columns from the summary for one reason or another

src/sensai/data_transformation.py
```diff
@@ -24,9 +25,14 @@ class DataFrameTransformer(ABC):
...
         self._name = f"{self.__class__.__name__}-{id(self)}"
         self.__isFitted = False
-        self._changesColumns = False
-        self.__removedColumns: Optional[Set[str]] = None
-        self.__addedColumns: Optional[Set[str]] = None
+        self._dfHistory: Optional[DataFrameHistoryTracker] = None
+
+    # for backwards compatibility with persisted DFTs based on code prior to commit 7088cbbe
+    # They lack the __isFitted attribute and we assume that each such DFT was fitted
+    def __setstate__(self, d):
```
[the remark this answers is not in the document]
MischaPanch: I removed the double underscores for this reason, see comment above. I could use `super().__setstate__` (Pycharm complained for some reason), but setting `__dict__` also works. We actually do that at one place already:

image

src/sensai/data_transformation.py
MischaPanch: Addressing your question:

>Why don't you think the history should be transient? It's only debugging data, effectively. Why would you want to retain >that when persisting an object? Especially now that you are retaining not only column names but also indices and not >just a pair but an arbitrarily long list (which I don't think is a good idea btw), this could become a lot of redundant data.

I will remove the indices. Concerning the column names change: typically a DFT will be persisted with a model. I want to be able to call the summary methods and to know what will happen in the entire data processing chain. Thus I want to keep the information about how the DFT changes the column names. With the way I implemented the history class, even after deleting the indices we do keep redundant information (columns before and after). But I don't think this can ever be a real downside in practical situations since it will occupy very little ram.

src/sensai/util/metadata.py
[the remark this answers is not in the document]
MischaPanch: I made it optional in the tracker, default False. We don't track indices in the DFTs anymore

src/sensai/data_transformation.py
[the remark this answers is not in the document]
MischaPanch: It's gone. If didn't want to persist the infos together with the object instances, I could have made the decorator useful by persisting the stuff not in the instance but in some global dict. But I did not want to make it transient so I saw no good solution.

Just out of curiosity, do you have an idea how to make it work as decorator without this ugly hack I did?

src/sensai/data_transformation.py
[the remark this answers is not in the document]
opcode81: What I meant was that `__str__` (or another method) could return "ColumnChange(<info>)" where <info> is, for example,
* "none" if there is no change
* "added=[...], removed=[...]" if columns are added or removed
* "reordered [new, order, of, columns]" otherwise

src/sensai/util/metadata.py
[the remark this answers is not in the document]
opcode81: Something is wrong with the way it is now. The history will only ever contain two entries, because the object is reset to a new object every time the decorated function is called.
Also, I can't think of a scenario where the same transformer instance would be used more than once in a single chain of transformations.

src/sensai/data_transformation.py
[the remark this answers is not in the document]
opcode81: Because changing the functionality of the method is not the only effect you want (you also want side-effects on the object's state), this simply isn't a use case for a function decorator.

opcode81 (changes requested):

src/sensai/featuregen.py
```diff
@@ -72,6 +72,11 @@ def __init__(self, categoricalFeatureNames: Optional[Union[Sequence[str], str]]
         self._name = None
         self.__isFitted = False
 
+    # for backwards compatibility with persisted Featuregens based on code prior to commit 7088cbbe
+    # They lack the __isFitted attribute and we assume that each such Featuregen was fitted
+    def __setstate__(self, d):
+        self.__isFitted = d.get("__isFitted", True)
```
opcode81: Also needs super call.

opcode81 (changes requested):

src/sensai/util/metadata.py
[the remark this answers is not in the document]
MischaPanch: For the decorator you are right but anyway it's gone now. The tracker is more general and I though it could be useful outside of the transformers, e.g. in an interactive setting. Say I have a bunch of feature generators and also do some stuff to a df myself in a python session. Then I can do:

```python
df = loadSomething()
tracker = DataFrameHistoryTracker(df)
doSomething(df)
tracker.update(df)
fgen1.apply(df)
tracker.update(df)
doSomethingElse(df)
tracker.update(df)
```

src/sensai/featuregen.py
MischaPanch: see comment above with the snippet from torch_base. For some reason pycharm complains about super().__setstate__, I was put off by that

src/sensai/util/metadata.py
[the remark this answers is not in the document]
opcode81: I don't think I would ever use this pattern. If I wanted to find out what's happening, I'd use the debugger rather than add such lines to my code.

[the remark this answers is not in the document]
opcode81: My suggestion is to do what is actually desired here: track column changes for DFTs.
Don't solve problems you don't have. Keep it simple, clear and to the point.

src/sensai/featuregen.py
opcode81: `super().__setstate__` is the most robust variant, because setting `self.__dict__` (which is the default implementation in object) is not necessarily OK if we later introduce a super-class.

opcode81 (changes requested):

src/sensai/data_transformation.py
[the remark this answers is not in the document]
MischaPanch: I pushed a change, is this what you meant?

src/sensai/data_transformation.py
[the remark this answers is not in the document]
MischaPanch: @opcode81 
Concerning
>super().__setstate__ is the most robust variant, because setting self.__dict__ (which is the default implementation in object) is not necessarily OK if we later introduce a super-class.

Pycharm was right to complain: when I replace it with `super().__setstate__(d)` the code does not run. 
```
    def __setstate__(self, d):
        d["_name"] = d.get("_name", f"{self.__class__.__name__}-{id(self)}")
        d["_isFitted"] = d.get("_isFitted", True)
        d["_dfHistory"] = d.get("_dfHistory", None)
>       super().__setstate__(d)
E       AttributeError: 'super' object has no attribute '__setstate__'
```
It seems like `ABC` is not derived from `object`, I didn't event know that was possible. Or there is some other black magic going on... Generally, it seems that in abstract classes `__setstate__` has to be implemented with the `__dict__` like I did it. Or am I missing something here?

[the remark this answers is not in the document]
opcode81: Not even `object` implements `__setstate__`. It's broken, which is unbelievable, really. The only classes I ever implemented `__setstate__` for were subclasses of `torch.nn.Module` (which implements `__setstate__`), so I didn't notice that `object` does not.
What this means is that when you implement `__setstate__` somewhere, you need to:
1. check if the super-class implements it and call it if so.
2. check if sub-classes implement it and if so change their implementations to call `super().__setstate__`.
This is insane! Everything would be totally "safe" if `object` implemented `__setstate__`; who's going to remember to do step 2?

Anyway, it's important to set the whole `__dict__` somehow (either via a call to super or explicitly), which you didn't previously, because you'd forget to update `__setstate__` when adding new properties.

src/sensai/data_transformation.py
```diff
@@ -181,6 +270,9 @@ def __init__(self, columns: Optional[Union[str, Sequence[str]]], categories: Uni
         :param ignoreUnknown: if True and an unknown category is encountered during transform, the resulting one-hot
             encoded columns for this feature will be all zeros. if False, an unknown category will raise an error.
         """
+        super().__init__()
+        self.columns = columns
```
opcode81, line 274: I really don't like the addition of new public members for informational purposes only, because
1. It's not immediately clear to even programmers of the class that the members are irrelevant for the functionality of the class (as the public status makes them appear more important than the actual members that do impact the functionality)
2. It's not clear to users of the class what the effect of changing the member is
3. (Minor) `summary` will raise an exception when called on previously persisted instances.

Why don't we keep all such parameters in a dictionary called, for example, `_consParams` or `_paramInfo`.

src/sensai/data_transformation.py
```diff
@@ -395,67 +496,111 @@ def apply(self, df: pd.DataFrame) -> pd.DataFrame:
         self._checkUnhandledColumns(df, matchedRulesByColumn)
         return df
 
+    def summary(self):
```
opcode81, line 499: Why are we calling the method `summary`? Wouldn't `info` be more a more obvious choice?

opcode81 (changes requested):

src/sensai/featuregen.py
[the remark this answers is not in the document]
opcode81: You didn't revert it, did you?

src/sensai/data_transformation.py
MischaPanch: I wanted to keep it similar to the keras interface. There models have the summary method. In the future I though of adding it to our models too

opcode81: Why? We are not keras.

src/sensai/featuregen.py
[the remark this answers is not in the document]
MischaPanch: I did, it logs on info now instead of raising an error

src/sensai/data_transformation.py
MischaPanch: At least I don't think it's worse than info. If you feel strongly about it, I can change it, of course

src/sensai/vector_model.py
```diff
@@ -189,13 +191,19 @@ def __init__(self, checkInputColumns=True):
...
-        self._isFitted = False
+        self._modelIsFitted = False
         self._predictedVariableNames = None
         self._modelInputVariableNames = None
         self._modelOutputVariableNames = ["UNKNOWN"]
         self._targetTransformer: Optional[InvertibleDataFrameTransformer] = None
         self.checkInputColumns = checkInputColumns
 
+    # for backwards compatibility with persisted models based on code prior to commit 7088cbbe
+    def __setstate__(self, d):
```
opcode81, line 202: Is this really necessary just for the property rename? Since the property is private, I don't think so.

opcode81 (changes requested):

src/sensai/util/metadata.py
```diff
@@ -0,0 +1,64 @@
...
+        Returns the columns in the first entry of the history that are not present in the last entry
+        """
+        return set(self.columnsHistory[0]).difference(self.columnsHistory[-1])
+
+    @lru_cache(maxsize=1)
+    def getAddedColumns(self):
+        """
+        Returns the columns in the last entry of the history that were not present the first one
+        """
+        return set(self.columnsHistory[0]).difference(self.columnsHistory[-1])
```
opcode81, line 51: This implements removed columns, not added columns!

src/sensai/util/metadata.py
```diff
@@ -0,0 +1,64 @@
...
+        self.indexHistory = [] if trackIndices else None
+        self.update(df)
+
+    def update(self, df: pd.DataFrame):
+        self.numUpdates += 1
+        self.columnsHistory.append(df.columns)
+        if self.trackIndices:
+            self.indexHistory.append(df.index)
+
+    @lru_cache(maxsize=1)
```
opcode81, line 39: Your cache causes the results to be incorrect if the function is called more than once on the same object.
You see, the sematics of `lru_cache` are to cache the result depending on the arguments. Since the only function argument is `self`, the cache will contain a cached value for one instance of the object. However, the hash code of the instance does not change when the object is updated (history item is added). Therefore the result will be wrong!
If it wasn't incorrect to start with, this would also be a premature optimisation, which you should *always* refrain from.

opcode81 (changes requested):

src/sensai/util/metadata.py
```diff
@@ -0,0 +1,64 @@
+from functools import lru_cache
```
opcode81, line 1: I think a better name for this module would be `pandas`, as it contains pandas-related utils (much like `utils.pickle` contains pickle-related utils).

src/sensai/featuregen.py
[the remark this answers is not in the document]
MischaPanch: Github diffs are far from perfect ^^. This is resolved then, right?

src/sensai/util/metadata.py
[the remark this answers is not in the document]
MischaPanch: I rewrote the class. I still called it tracker since I couldn't think of a better name for the method taking the transformed DF. If you have a better suggestion, I am happy to change the naming

src/sensai/data_transformation.py
MischaPanch: added `_paramInfo` to the base class and removed columns from the public interface

src/sensai/data_transformation.py
MischaPanch: It really is more of an info than a summary. I changed the name

src/sensai/featuregen.py
```diff
@@ -194,17 +223,22 @@ class RuleBasedFeatureGenerator(FeatureGenerator, ABC):
...
     """
-    def __init__(self, featureGenerators: Sequence[FeatureGenerator]):
-        """
-        :param featureGenerators:
-        """
-        self.featureGenerators = featureGenerators
+    def __init__(self, *featureGenerators: Union[FeatureGenerator, List[FeatureGenerator]]):
+        self.featureGenerators = flattenArguments(featureGenerators)
+        if len(self.featureGenerators) == 0:
+            log.info(f"MultiFeatureGenerator instance {self.getName()} is empty")
```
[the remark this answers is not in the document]
MischaPanch: Nice catch, I really shouldn't write code at 2am... Adjusted the log message

src/sensai/featuregen.py
```diff
@@ -104,6 +122,9 @@ def getGeneratedColumnNames(self) -> Optional[List[str]]:
         return self._generatedColumnNames
 
     @abstractmethod
+    def _fit(self, X: pd.DataFrame, Y: pd.DataFrame = None, ctx=None):
+        pass
```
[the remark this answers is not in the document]
MischaPanch: Done

src/sensai/featuregen.py
```diff
@@ -674,6 +754,11 @@ def _generate(self, df: pd.DataFrame, ctx=None) -> pd.DataFrame:
             log.info(f"Generating target features via {self.vectorModel.__class__.__name__}")
             return self.vectorModel.predict(df)
 
+    def summary(self):
+        summary = super().summary()
+        summary["wrappedModelName"] = self.vectorModel.getName()
```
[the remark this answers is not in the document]
MischaPanch: True, I changed it

src/sensai/vector_model.py
MischaPanch: Yeah, not worth the additional complexity. I Reverted the change in name it and added a block comment to the _isFitted instead.

src/sensai/util/metadata.py
MischaPanch: Fixed

src/sensai/util/metadata.py
MischaPanch: Thanks!, I had a fundamental misunderstanding there. Removed it

src/sensai/util/metadata.py
MischaPanch: Renamed it

src/sensai/util/metadata.py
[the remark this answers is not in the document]
opcode81: The name is fine, though I would go with the full "DataFrame" prefix; we don't need to save letters here.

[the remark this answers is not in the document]
MischaPanch: Done. I think all comments were addressed now and we can merge this

src/sensai/data_transformation.py
```diff
@@ -181,6 +272,9 @@ def __init__(self, columns: Optional[Union[str, Sequence[str]]], categories: Uni
         :param ignoreUnknown: if True and an unknown category is encountered during transform, the resulting one-hot
             encoded columns for this feature will be all zeros. if False, an unknown category will raise an error.
         """
+        super().__init__()
+        self._paramInfo["columns"] = columns
+        self.inferCategories = categories is None
```
opcode81, line 277: This should be added to _paramInfo and "categories is not None" should be used below

src/sensai/data_transformation.py
```diff
@@ -224,37 +318,51 @@ def apply(self, df: pd.DataFrame):
                 df["%s_%d" % (columnName, i)] = encodedArray[:, i]
         return df
 
+    def info(self):
+        info = super().info()
+        info["columns"] = self._paramInfo["columns"]
```
opcode81, line 323: Consider using `info.update(self._paramInfo)` for easy extensibility

src/sensai/featuregen.py
```diff
@@ -82,6 +91,15 @@ def getName(self) -> str:
     def setName(self, name):
         self._name = name
 
+    def info(self):
+        return {
+            "name": self.getName(),
+            "categoricalFeatureNames": self.__categoricalFeatureNames,
+            "generatedColumnNames": self.getGeneratedColumnNames(),
+            "isFitted": self.isFitted(),
+            "normalizationRules": self.getNormalisationRules(),
```
opcode81, line 100: AmE

opcode81 (approved):
Only a few very minor things remain. Can be merged after they've been addressed.

src/sensai/featuregen.py
MischaPanch: Done

src/sensai/data_transformation.py
MischaPanch: Done

src/sensai/data_transformation.py
MischaPanch: Done
Aaaand merging this :)

Issue #34, opened 2020-12-09
Maybe improve _predict interface in VectorModel


MischaPanch:
To make this "cleaner", we could consider changing the interface of `_predict` to not return a DF but a more low-level result instead and make `predict` construct the DF correctly.

_Originally posted by @opcode81 in https://github.com/jambit/sensAI/pull/33#discussion_r539355532_

opcode81:
Implements should be allowed to return either a pd.DataFrame, a pd.Series (if it's a one-column output) or an np.array - and VectorModel should take care of constructing the proper DF from it (and ensure that it has the right index).

opcode81:
Has been addressed with commit 5002840360c22adf1a0046cc73907bd0a26f6536

PR #42, opened 2021-01-20
Improved hierarchy in vector_model, extended evaluation

MischaPanch:
This PR introduces a "more relaxed" hierarchy in vector_model, making it easier to get functionality by inheriting from selected classes.

There are also some functional improvements and enhancements, see the commit messages for that. The more fine-granular methods for pre/post processing and input validation allow for better unit testing, which I have not implemented yet.

One important functional change is the added default implementation of _predict to VectorClassificationModel. Since I potentially will be inheriting from this class a lot, it reduces code duplication (and complexity of implementation).

Another one is that I removed postprocessing from the PredictorModel base class. I figured this functionality is not needed for rule based models

I will make some comments on noteworthy places to simplify the review for you.


src/sensai/torch/torch_models.py
```diff
@@ -148,8 +147,8 @@ def _predictOutputsForInputDataFrame(self, inputs: pd.DataFrame) -> np.ndarray:
         result = super()._predictOutputsForInputDataFrame(inputs)
         return np.squeeze(result, 2)
 
-    def _computeInputs(self, x: pd.DataFrame, y=None, fit=False) -> pd.DataFrame:
-        x = super()._computeInputs(x, Y=y, fit=fit)
+    def computeProcessedInputs(self, x: pd.DataFrame, Y: pd.DataFrame = None, fit=False) -> pd.DataFrame:
+        x = super().computeProcessedInputs(x, Y=Y, fit=fit)
```
MischaPanch, line 151: I am a little bit unsure whether this change is correct (like 90% sure it's fine, but still). It might deserve a critical look from you

src/sensai/vector_model.py
```diff
@@ -33,13 +34,44 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
...
         pass
 
     @abstractmethod
-    def isRegressionModel(self) -> bool:
+    def getPredictedVariableNames(self) -> list:
         pass
 
+    def fitPreprocessors(self, X: pd.DataFrame, Y: pd.DataFrame = None):
+        # no need for fitGenerate if chain is empty
+        if len(self._inputTransformerChain) == 0 and self._featureGenerator is not None:
```
MischaPanch, line 46: There was a mistake in RuleBased model before, where the chain was not fitted on the generator's output. The new method fixes it. Depending on the nature of the feature generator, treating the common case of trivial DFTchains separately can result in a significant improvement in performance

src/sensai/vector_model.py
```diff
@@ -33,13 +34,44 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
...
+        :param fit: if True, preprocessors will be fitted before being applied to X
+        :return: 
+        """
+        if fit:
+            if self._featureGenerator is not None:
+                X = self._featureGenerator.fitGenerate(X, Y, self)
+            X = self._inputTransformerChain.fitApply(X)
+        else:        
+            if Y is not None:
+                log.warning(f"Passed an output data frame Y although preprocessors are already fitted. "
```
MischaPanch, line 68: I figured it's not worth an exception, what do you think?

src/sensai/vector_model.py
```diff
@@ -161,39 +163,61 @@ def isFitted(self) -> bool:
...
 
-class RuleBasedModel(FittableModel, ABC):
-    """
-    Base class for models where the essential prediction logic is based on rules coded by humans
-    and thus does not require fitting. However, the input generation process may use mechanisms
-    such as feature generation or data transformation, which may require fitting.
-    """
-    def __init__(self):
-        super().__init__()
+class ClassificationMixin(ABC):
```
MischaPanch, line 166: Had to do a mixin because RuleBasedClassificationModel cannot inherit from the corresponding vector model (for performance reasons). The mixin might also be useful in other situations, I think

src/sensai/vector_model.py
```diff
@@ -161,39 +163,61 @@ def isFitted(self) -> bool:
...
+        if list(predictionDf.columns) != labels:
+            raise Exception(f"{self} _predictClassProbabilities returned DataFrame with incorrect columns: "
+                            f"expected {labels}, got {predictionDf.columns}")
 
-        :param x: the input data
-        :return: a DataFrame with the same index as the input
+        dfToCheck = predictionDf.iloc[:maxRowsToCheck]
+        for i, (_, valueSeries) in enumerate(dfToCheck.iterrows(), start=1):
+
+            if not all(0 <= valueSeries) or not all(valueSeries <= 1):
```
MischaPanch, line 194: This is an enhancement of the previous check

src/sensai/vector_model.py
```diff
@@ -360,41 +405,102 @@ def _fitClassifier(self, X: pd.DataFrame, y: pd.DataFrame):
...
+    def _predict(self, X: pd.DataFrame) -> pd.DataFrame:
+        pass
+
+
+class RuleBasedRegressionModel(RuleBasedModel, ABC):
+    def isRegressionModel(self) -> bool:
+        return True
+
+
+class RuleBasedClassificationModel(ClassificationMixin, RuleBasedModel, ABC):
```
MischaPanch, line 473: There is some code duplication between here and VectorClassificationModel. I could reuse some stuff with the mixin but not all. I think it is bearable and didn't find a better way of organizing the classes. If you have a suggestion for how to improve it, please shoot! :)

src/sensai/vector_model.py
opcode81: This is becoming overly complex. I shall make a proposal (via a PR) on how to simplify things.
Please don't make further changes to vector_model until I'm done.

MischaPanch: Awesome, thank you!

src/sensai/vector_model.py
```diff
@@ -33,13 +34,44 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
...
+    def fitPreprocessors(self, X: pd.DataFrame, Y: pd.DataFrame = None):
+        # no need for fitGenerate if chain is empty
+        if len(self._inputTransformerChain) == 0 and self._featureGenerator is not None:
+            self._featureGenerator.fit(X, Y)
+
+        elif self._featureGenerator is not None:
+            X = self._featureGenerator.fitGenerate(X, Y, self)
+        self._inputTransformerChain.fit(X)
+
+    def computeProcessedInputs(self, X: pd.DataFrame, Y: pd.DataFrame = None, fit=False) -> pd.DataFrame:
```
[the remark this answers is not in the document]
MischaPanch: Yes, let's do it like this. The non-fitting part is a useful part of the public interface, I think (e.g. for debugging in interactive sessions)

src/sensai/vector_model.py
```diff
@@ -33,13 +34,44 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
...
-    def getPredictedVariableNames(self):
+    def isRegressionModel(self) -> bool:
         pass
 
     @abstractmethod
-    def isRegressionModel(self) -> bool:
+    def getPredictedVariableNames(self) -> list:
         pass
 
+    def fitPreprocessors(self, X: pd.DataFrame, Y: pd.DataFrame = None):
```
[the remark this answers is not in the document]
MischaPanch: I see your point, let's make it private then.

src/sensai/vector_model.py
MischaPanch: Outdated

src/sensai/vector_model.py
```diff
@@ -389,7 +406,6 @@ def convertClassProbabilitiesToPredictions(self, df: pd.DataFrame):
         """
         modelOutputVariableName = self.getModelOutputVariableNames()[0]
         y = self._convertClassProbabilitiesToPredictions(df, predictedVariableName=modelOutputVariableName)
-        # TODO check sanity: where should we be applying post-processing?
```
opcode81: image

opcode81: You fixed this by calling `_convertClassProbabilities` instead, but I would argue that it is rather unexpected for this method to do anything but what its name suggests, which should not include the post-processing. This is also a breaking change, because the previous public method did not apply post-processing.
Therefore, I would propose to keep only the public method (and inline the private method - as it was before).

src/sensai/vector_model.py
```diff
@@ -125,6 +128,10 @@ def withTargetTransformer(self, targetTransformer: Optional[InvertibleDataFrameT
             the model, i.e. the transformation is completely transparent when applying the model.
         :return: self
         """
+        # Note: it is important to disallow targetTransformers for rule based models since we need
+        # predictedVarNames and modelOutputVarNames to coincide there.
```
opcode81, line 132: I don't understand this reasoning. The use of a transformer doesn't prevent them from being the same, and why do we need them to be the same?
In my mind, the reason why it is non-sensical to apply a target transformer is that since you are not training, you don't actually use any targets. And during inference, you obviously generate the targets yourself, so you could really only ever use the inverse (which an output transformer would be the better solution for).

src/sensai/vector_model.py
```diff
@@ -110,12 +109,16 @@ def withOutputTransformers(self, *outputTransformers: Union[DataFrameTransformer
...
+        # There is no reason for post processing in rule based models.
+        if not self._underlyingModelRequiresFitting():
+            raise Exception(f"Output transformers are not supported for model of type {self.__class__.__name__}")
         self._outputTransformerChain = DataFrameTransformerChain(*outputTransformers)
         return self
 
     def withTargetTransformer(self, targetTransformer: Optional[InvertibleDataFrameTransformer]) -> __qualname__:
         """
-        Makes the model use the given target transformers.
+        Makes the model use the given target transformers. Not supported for rule based models.
```
opcode81, line 121: Use hyphen in "rule-based" (below also)

opcode81:
Please use more detailed commit messages (not just in this PR, but in general). For example, "Fixed bug in XY" or "Improvements in XY" does not contain a lot of information. What was the bug, how was it fixed, what is the improvement?

src/sensai/evaluation/__init__.py
```diff
@@ -3,4 +3,4 @@
 from .eval_util import RegressionEvaluationUtil, ClassificationEvaluationUtil, MultiDataEvaluationUtil, \
     evalModelViaEvaluator, createEvaluationUtil, createVectorModelEvaluator, createVectorModelCrossValidator
 from .evaluator import VectorClassificationModelEvaluator, VectorRegressionModelEvaluator, \
-    VectorRegressionModelEvaluationData, VectorClassificationModelEvaluationData
```
opcode81, line 6: Why have so many classes in the evaluation modules been renamed?

opcode81 (changes requested):

src/sensai/vector_model.py
opcode81: Could come in handy, for sure.

src/sensai/vector_model.py
MischaPanch: Apart from being nonsensical for rule based models (which I could have tolerated), it is actively damaging there. The modelOutputVars getter is used in VectorClassification, so it is crucial to get the names right. The predictedVars are provided by the user for rule-based models, not so the modelOutputVars.

The only difference between modelOutputVars and predictedVars is due to a possible pass through the target transformer. So I don't understand your statement - disallowing target transformers for rule-based model forces the two notions to agree and solves the problem described above

src/sensai/evaluation/__init__.py
MischaPanch: As a preparation for tensor model - I will use this classes in the TensorModelEvaluator. If you want, I can revert the change for the time being. The EvaluationData classes seemed to me more general than VectorModel, therefore the renaming

MischaPanch: If you want, I can revert this for the time being

MischaPanch: In preparation for tensor model - I will base the evaluation classes for tensor model from ModelEvaluationData

> Please use more detailed commit messages (not just in this PR, but in general). For example, "Fixed bug in XY" or "Improvements in XY" does not contain a lot of information. What was the bug, how was it fixed, what is the improvement?

Right, I got sloppy. I will reword them a bit

src/sensai/vector_model.py
MischaPanch: The previous public method was incorrect - it is used in the evaluators and would lead to evaluating the non-processed predictions instead of the actual model. The current version of the code fixes this bug.

I would argue that it is expected to get the non-processed results. In fact, bevor this PR the dosctring of the method (falsely) claimed: "...to a result as return by predict". I will adjust the docstring to the original version in my next commit, forgot to do it before.

Here is the previously faulty peace of code that is fixed now:

image

src/sensai/vector_model.py
```diff
@@ -358,43 +377,127 @@ def getClassLabels(self) -> List[Any]:
...
     def _predictClassProbabilities(self, X: pd.DataFrame) -> pd.DataFrame:
         pass
+
+    def _predict(self, x: pd.DataFrame) -> pd.DataFrame:
+        modelOutputVariableName = self.getModelOutputVariableNames()[0]
+        predictedProbabilitiesDf = self._predictClassProbabilities(x)
+        return self._convertClassProbabilitiesToPredictions(predictedProbabilitiesDf, predictedVariableName=modelOutputVariableName)
+
+
+class RuleBasedRegressionModel(VectorRegressionModel, ABC):
```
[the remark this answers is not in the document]

src/sensai/vector_model.py
opcode81: The reasoning is illogical, because the use of a target transformer does not prevent them from being the same, like I said, and the only reason you need them to coincide is because you have modified `getModelOutputVariableNames` to force this (and perhaps just setting the member at construction - with an explanation that target transformers are disallowed and therefore they coincide - would be clearer).
The real reason why it should not be allowed to use target transformers the one I state in my second paragraph above.
Your reasoning in the comment we are talking about is due to a self-inflicted problem and therefore does not make sense without reading all the code.

src/sensai/vector_model.py
```diff
@@ -358,43 +377,127 @@ def getClassLabels(self) -> List[Any]:
...
+        dfToCheck = predictionDf.iloc[:maxRowsToCheck]
         for i, (_, valueSeries) in enumerate(dfToCheck.iterrows(), start=1):
+
+            if not all(0 <= valueSeries) or not all(valueSeries <= 1):
+                log.warning(f"Probabilities data frame may not be correctly normalised, "
+                            f"got probabilities outside the range [0, 1]: checked row {i}/{maxRowsToCheck} contains {list(valueSeries)}")
+
             s = valueSeries.sum()
-            if abs(s-1.0) > 0.01:
+            if not np.isclose(s, 1):
```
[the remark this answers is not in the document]
MischaPanch: Reduced to original tolerance. Although this only logs an error, so we could tolerate higher tolerance ^^ (pun intended)

src/sensai/vector_model.py

src/sensai/vector_model.py
```diff
@@ -214,10 +214,16 @@ def _checkModelInputColumns(self, modelInput: pd.DataFrame):
...
-    def computeProcessedInputs(self, X: pd.DataFrame, Y: pd.DataFrame = None, fit=False) -> pd.DataFrame:
+    def computeModelInputs(self, X):
         """
         Returns the dataframe that is passed to the model, i.e. the result of applying preprocessors to X.
+        """
+        if self._featureGenerator is not None:
+            X = self._featureGenerator.generate(X, self)
+        return self._inputTransformerChain.apply(X)
 
+    def _computeProcessedInputs(self, X: pd.DataFrame, Y: pd.DataFrame = None, fit=False) -> pd.DataFrame:
```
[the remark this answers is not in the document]
MischaPanch: Yeah, much better this way, I changed it

src/sensai/vector_model.py
opcode81: The error is not in `convertClassProbabilitiesToPredictions` but in this evaluator, which does not take into account that there might be an output transformer.
The phrase "a result as returned by predict" does not mean that it is *the* same result but *a* result of the same form.

src/sensai/vector_model.py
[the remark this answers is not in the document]
opcode81: True. Haha.

src/sensai/vector_model.py
opcode81: But the issue you found again makes me question whether we should be keeping output transformers.
And target transformers probably make sense only for regressors.

src/sensai/vector_model.py
opcode81: Long story short, I would propose 
* to change the comment to something like `# Disabled for rule-based models which do not apply fitting and therefore cannot make use of transformed targets`
* Explicitly set `_modelOuputVariableNames` at construction with the remark `# guaranteed to be the same as predictedVariableNames since target transformers and output transformers are disallowed`

src/sensai/evaluation/__init__.py
opcode81: > If you want, I can revert this for the time being

If you can easily revert, I would prefer this until we are ready to reassess the needs as new types of models are implemented.

src/sensai/vector_model.py
MischaPanch: I am not a big fan of the output transformers myself, as you know. A quick, non-invasive fix for the issue in the above evaluator would be to move all post processing from the base class to VectorRegressionModel.

If we don't do that, fixing the evaluator would be a bit hacky. _applyPostProcessors is private (and I don't want to make it public) so we would need to recreate it's functionality in the evaluator. This code duplication seems too fragile to me

src/sensai/vector_model.py
MischaPanch: Right, I got what you meant now. Adjusted the comment and set the field at construction

src/sensai/vector_model.py
MischaPanch: @opcode81 I went ahead and moved the post-processing to regression models only. The code seems clearer to me now. This change is in a single commit so it can easily be reverted, in case you disagree

src/sensai/vector_model.py
```diff
@@ -291,32 +213,16 @@ def fit(self, X: pd.DataFrame, Y: Optional[pd.DataFrame], fitPreprocessors=True,
...
+            log.info(
+                f"Training with outputs[{len(Y.columns)}]={list(Y.columns)}, inputs[{len(self._modelInputVariableNames)}]=[{', '.join([n + '/' + X[n].dtype.name for n in self._modelInputVariableNames])}]")
             self._fit(X, Y)
             self._isFitted = True
 
     @abstractmethod
     def _fit(self, X: pd.DataFrame, Y: pd.DataFrame):
         pass
 
-    def getModelOutputVariableNames(self):
```
opcode81, line 318: Look for references to this functions; you'll probably find a few dead ones in classification models.
I am on the fence as to whether this method should exist in VectorModel and return predictedVariablesNames or make this a breaking change and only fix it in sensai itself for the time being.

opcode81 (commented):
In response to 
"@opcode81 I went ahead and moved the post-processing to regression models only. The code seems clearer to me now. This change is in a single commit so it can easily be reverted, in case you disagree":

I would have proposed to the same. It was the right thing to do.
With regression models, the output transformers are not as bad, so we can keep them around for longer. You missed your chance of getting rid of them. ;)

src/sensai/vector_model.py
MischaPanch: I'll have a look within sensai. I would find this method confusing in the classifier's public interface, given that post-processing is gone there now

opcode81: Improvement #34 will remove the need for implementations to use any of these functions.

opcode81: I will implement it one of these days; I'd like to.
So for now, just change it to use predictedVarNames.

MischaPanch: Found one and fixed it. You have a good "Riecher", I didn't even think about that

opcode81 (approved):

PR #43, opened 2021-01-20
Restore class hierarchy

opcode81:
The **first commit only reorders code**, putting functions back into the place where they were before the introduction of rule-based models (compare with commit 7989de7d if you're interested in what changed compared to that revision).
The commit does not change *any* actual lines of code; I only added some TODOs.

Basically, you should necessarily review only the second commit.
And we need to discuss the TODOs that are still there.

### DIFF
--- a/src/sensai/vector_model.py
+++ b/src/sensai/vector_model.py
@@ -19,14 +19,11 @@ from .util.sequences import getFirstDuplicate
 log = logging.getLogger(__name__)
 
 
-class PredictorModel(PickleLoadSaveMixin, ABC):
+class PredictorModel(ABC):
     """
     Base class for models that map data frames to predictions
     """
     def __init__(self):
-        self._featureGenerator: Optional[FeatureGenerator] = None
-        self._inputTransformerChain = DataFrameTransformerChain()
-        self._outputTransformerChain = DataFrameTransformerChain()
         self._name = None
 
     @abstractmethod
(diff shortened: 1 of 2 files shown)


src/sensai/vector_model.py
```diff
@@ -334,6 +247,23 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
     def _predict(self, x: pd.DataFrame) -> pd.DataFrame:
         pass
 
+    def _underlyingModelRequiresFitting(self) -> bool:
+        """
+        Designed to be overridden for rule-based models.
+
+        :return: True iff the underlying model requires fitting
```
MischaPanch, line 254: though I like mathematical notation, here "if" with one f will be enough

src/sensai/vector_model.py
```diff
@@ -434,73 +431,27 @@ def _predict(self, x: pd.DataFrame) -> pd.DataFrame:
         return self.convertClassProbabilitiesToPredictions(self._predictClassProbabilities(x))
 
 
-class RuleBasedModel(FittableModel, ABC):
-    """
-    Base class for models where the essential prediction logic is based on rules coded by humans
-    and thus does not require fitting. However, the input generation process may use mechanisms
-    such as feature generation or data transformation, which may require fitting.
-    """
+class RuleBasedRegressionModel(VectorRegressionModel):
```
MischaPanch, line 434: The structure is much simpler this way, thank you! What bugs me a bit is that rule based models now can have target transformer and output transformer. I can't really think of a case where it is ever useful. A hacky way to resolve this would be to override those methods as well, causing them to raise exceptions or log errors (the code would have to be duplicated too)

We could, of course, also take the approach to let the users think for themselves...

src/sensai/vector_model.py
opcode81: Not really. If one "f" is removed, then the specification is incomplete. For it to be complete, one would need to add ", False otherwise".
Why do you want me to change this?

src/sensai/vector_model.py
opcode81: I don't understand why that bugs you.
Having options is a good thing.

src/sensai/vector_model.py
```diff
@@ -376,7 +389,7 @@ def convertClassProbabilitiesToPredictions(self, df: pd.DataFrame):
         """
         modelOutputVariableName = self.getModelOutputVariableNames()[0]
         y = self._convertClassProbabilitiesToPredictions(df, predictedVariableName=modelOutputVariableName)
-        # TODO check sanity
+        # TODO check sanity: where should we be applying post-processing?
```
MischaPanch, line 392: The way it is now, post-processing can only happen after the argmax operation - the invertible target transformer can convert between labels and back. We can't feed probability vectors to it

src/sensai/vector_model.py
MischaPanch: More options lead to more room for errors (especially in design), that's why I mentioned it. I guess I worry too much about this anyway, probably no-one will even get the idea of using target transformer for rule based stuff

src/sensai/vector_model.py
```diff
@@ -39,7 +39,7 @@ def getPredictedVariableNames(self) -> list:
         pass
 
 
-# TODO Is this obsolete?
+# TODO Do we need this class?
```
MischaPanch, line 42: The only reason to keep it could be for type annotations in the evaluation module. The base evaluator classes only need the predictor and fittable model interfaces. Which means if we at some point want fittable models that are not vector models (or we want to allow our advanced users to build them), the base evaluator classes can be used for them as well

Do you think it's enough of a reason? It probably doesn't hurt to keep it, I can't see any downsides

src/sensai/vector_model.py
```diff
@@ -379,13 +312,39 @@ def getModelOutputVariableNames(self):
     def getPredictedVariableNames(self):
```
MischaPanch, line 312: This can give incorrect results in the rule based models. Could be settled by setting `_predictedVariableNames` in the rule based models' constructors

src/sensai/vector_model.py
opcode81: It's true that some of these options don't make sense to use, but I think it would be highly unlikely for users to make use of them. So I really wouldn't worry.

src/sensai/vector_model.py
MischaPanch: Same for `_modelOutputVariableNames `, which for rule based models should coincide (if the user did't come up with the idea of using a name-changing target transformer). I would say we don't need this variable, but we have the getter as part of the public interface

MischaPanch: Ah, maybe there is an acceptable way around: use the new `_underlyingModelRequiresFitting` flag to give a warning or even raise an exception in the getters. The docu should then say: not supported for rule based models.

opcode81: It will be set if users call `fit`, which could be appropriate, since there may be parts that still require fitting. 
We could solve it by extending the documentation of the method ("will return None if `fit` has not been called").

src/sensai/vector_model.py
opcode81: Yeah, let's keep it.

src/sensai/vector_model.py
```diff
@@ -334,6 +247,23 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
...
+    def fitPreprocessors(self, X: pd.DataFrame, Y: pd.DataFrame = None):
+        # no need for fitGenerate if chain is empty
+        if self._featureGenerator is not None:
+            if len(self._inputTransformerChain) == 0:
+                self._featureGenerator.fit(X, Y)
+            else:
+                X = self._featureGenerator.fitGenerate(X, Y, self)
+        self._inputTransformerChain.fit(X)
+
     def fit(self, X: pd.DataFrame, Y: pd.DataFrame, fitPreprocessors=True, fitTargetTransformer=True):
```
MischaPanch, line 267: I have to be able to fit rule based models without passing Y. Would it be ok for you to make Y None by default? Otherwise the user has to remember to explicitly pass None. Docu for Y also has to be adjusted

src/sensai/vector_model.py
opcode81: I like your proposal. Make this getter log a warning if the value is None and the flag is true.
The other getter (`getModelOutputVariableNames`) should raise an exception.

src/sensai/vector_model.py
opcode81: You can't do this in VectorModel because it would suggest an incorrect interface to users.
You could do it in rule-based subclasses though.

src/sensai/vector_model.py
MischaPanch: `_predictedVariableNames` won't always be set in fit for rule based models (see my comment for fit below) - Y can be None

src/sensai/vector_model.py
MischaPanch: Sounds good. Anyway fit requires some special docu in the rule based stuff, this gives the possibility to add it

src/sensai/vector_model.py
```diff
@@ -233,13 +62,27 @@ def __init__(self, checkInputColumns=True):
...
         super().__init__()
+        self._featureGenerator: Optional[FeatureGenerator] = None
+        self._inputTransformerChain = DataFrameTransformerChain()
+        self._outputTransformerChain = DataFrameTransformerChain()
         self._isFitted = False  # Note: this keeps track only of the actual model being fitted, not the pre/postprocessors
         self._predictedVariableNames: Optional[list] = None
         self._modelInputVariableNames: Optional[list] = None
         self._modelOutputVariableNames: Optional[list] = None
         self._targetTransformer: Optional[InvertibleDataFrameTransformer] = None
+        # TODO this was recently (December 6, 2020) introduced and will cause compatibility issues; should be removed
```
MischaPanch, line 73: not sure what you mean, `checkInputColumns` was there before the date you mention. I don't remember adding it myself (though I might well have, who knows ^^)

src/sensai/vector_model.py
MischaPanch: Hm, we could use the same trick as with the columnNameGetters here: raise an exception if the `withTarget/OutputTransformer` is called and the underlying model does not require fitting - as an indication that it doesn't make sense to use them

src/sensai/vector_model.py
opcode81: No, it's older. I thought it was older but was fooled by the "blame" output. :)

opcode81: You probably added this when you introduced the ensemble models.
It struck me as odd because it's the only non-private property of VectorModel.

src/sensai/vector_model.py
opcode81: True. This dilutes the interface of `fit` somewhat, because, strictly speaking, we would need to annotate Y with `Optional[DataFrame]`. In Java, we wouldn't have this problem. :-/

src/sensai/vector_model.py
opcode81: If you look at the implementation of `VectorModel.predict`, you'll find that it's being applied at the very end, after having called `_predict`, which, for classification models, now (indirectly) calls `_applyPostProcessing` again in `convertClassProbabilitiesToPredictions`.
I see a problem there. ;)

Anyway, this will be my last comment. I won't have time to take this any further. If you want to, take over.

src/sensai/vector_model.py
opcode81: It just occurred to me that evaluation makes use of `getPredictedVariableNames` so we really need to get these names one way or another.

PR #46, opened 2021-01-27
Feature/towards tensor model

### LINKED ISSUE #14: Support for tensor-like output

MischaPanch:
Closes #14 

At the current stage, this is an empty hull. I will add a notebook implementing some tensor to scalar and tensor to tensor models. The evaluation classes for tensor to tensor will be added in a future PR (by a colleague of mine at apliedAI)

I mainly wanted to get your opinion on the design choices I made

EDIT: Added notebook and pytorch-lightning wrappers. I believe that many new kinds of models can now easily be wrapped by sensai classes

### DIFF
--- a/src/sensai/data_ingest.py
+++ b/src/sensai/data_ingest.py
@@ -79,11 +79,8 @@ TInputOutputData = TypeVar("TInputOutputData", bound=BaseInputOutputData)
 
 class DataSplitter(ABC):
     @abstractmethod
-    def split(self, data: TInputOutputData) -> Tuple[InputOutputDataFrames, InputOutputDataFrames]:
+    def split(self, data: TInputOutputData) -> Tuple[TInputOutputData, TInputOutputData]:
         pass
-
-
-IOData = TypeVar("IOData", bound=InputOutputData)
 
 
 class DataSplitterFractional(DataSplitter):
@@ -94,7 +91,7 @@ class DataSplitterFractional(DataSplitter):
         self.shuffle = shuffle
         self.randomSeed = randomSeed
 
-    def split(self, data: IOData) -> Tuple[IOData, IOData]:
+    def split(self, data: TInputOutputData) -> Tuple[TInputOutputData, TInputOutputData]:
         numDataPoints = len(data)
         splitIndex = int(numDataPoints * self.fractionalSizeOfFirstSet)
         rand = np.random.RandomState(self.randomSeed)
--- a/src/sensai/data_ingest.py
+++ b/src/sensai/data_ingest.py
@@ -1,13 +1,17 @@
 from abc import ABC, abstractmethod
-from typing import Tuple, Sequence
+from typing import Tuple, Sequence, TypeVar
 
 import numpy as np
 import pandas as pd
 import scipy.stats
 
 
-class InputOutputData:
-    def __init__(self, inputs: pd.DataFrame, outputs: pd.DataFrame):
+class BaseInputOutputData(ABC):
+    def __init__(self, inputs, outputs):
+        """
+        :param inputs: expected to have shape and __len__
+        :param outputs: expected to have shape and __len__
+        """
         if len(inputs) != len(outputs):
             raise ValueError("Lengths do not match")
         self.inputs = inputs
@@ -16,6 +20,40 @@ class InputOutputData:
     def __len__(self):
         return len(self.inputs)
 
+    @abstractmethod
+    def filterIndices(self, indices: Sequence[int]) -> __qualname__:
+        pass
+
+
+class InputOutputArrays(BaseInputOutputData):
+    def __init__(self, inputs: np.ndarray, outputs: np.ndarray):
+        super().__init__(inputs, outputs)
+
+    def filterIndices(self, indices: Sequence[int]) -> __qualname__:
+        inputs = self.inputs[indices]
+        outputs = self.outputs[indices]
+        return InputOutputArrays(inputs, outputs)
+
+    def toTorchDataLoader(self, batchSize=64, shuffle=True):
+        try:
+            import torch
+            from torch.utils.data import DataLoader, TensorDataset
+        except ImportError:
+            raise ImportError(f"Could not import torch, did you install it?")
+        dataSet = TensorDataset(torch.tensor(self.inputs), torch.tensor(self.outputs))
+        return DataLoader(dataSet, batch_size=batchSize, shuffle=shuffle)
+
+
+# TODO: Rename to InputOutputDataFrames when the time for breaking changes has come
+class InputOutputData(BaseInputOutputData):
+    def __init__(self, inputs: pd.DataFrame, outputs: pd.DataFrame):
+        super().__init__(inputs, outputs)
+
+    def filterIndices(self, indices: Sequence[int]) -> __qualname__:
+        inputs = self.inputs.iloc[indices]
+        outputs = self.outputs.iloc[indices]
+        return InputOutputData(inputs, outputs)
+
     @property
     def inputDim(self):
         return self.inputs.shape[1]
@@ -23,11 +61,6 @@ class InputOutputData:
     @property
     def outputDim(self):
         return self.outputs.shape[1]
-
-    def filterIndices(self, indices: Sequence[int]) -> 'InputOutputData':
-        inputs = self.inputs.iloc[indices]
-        outputs = self.outputs.iloc[indices]
-        return InputOutputData(inputs, outputs)
 
     def computeInputOutputCorrelation(self):
         correlations = {}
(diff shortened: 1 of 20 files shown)


opcode81:
Unfortunately, there is no notebook support in PyCharm Community, so I need to find another way of taking a look at the notebooks.

opcode81:
General observation: It seems that you are very concerned with avoiding code duplication. However, factorisation is _not_ more important than other aspects of the design, e.g. clarity or maintainability. This is very important, as you seem to be willing to sacrifice virtually anything to avoid code duplication, which is not a healthy maxim at all.

opcode81 (changes requested):

src/sensai/tensor_model.py
```diff
@@ -0,0 +1,392 @@
...
+def _checkDfShape(df: pd.DataFrame, desiredShape: tuple):
+    datapointShape = _getDatapointShape(df)
+    if datapointShape != desiredShape:
+        raise InvalidShapeError(f"Wrong input shape for data point. Expected {desiredShape} but got {datapointShape}")
+
+
+# This has to be implemented as a mixin because there can be no functional common class for tensor models.
+# The reason is that actual implementations need to inherit from Vector-Regression/Classification-Model
+# (or duplicate a lot of code) and thus it is not possible to inherit from something like TensorModel(VectorModel)
+# without getting into a mess. Despite that, we want things to "be a TensorModel", hence this class.
```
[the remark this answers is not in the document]
MischaPanch: I had two reasons in mind for TensorModel base classes:

1) (the less important one) - Normal vector model evaluators will not work with tensor models (which are also VectorModels). So I thought having separate classes and evaluators for them would make things a bit easier. We could use stuff like isinstance(TensorModel) for catching user errors

2) Better user experience. An important use case for tensor models will be that users implement them. Currently, only two methods need to be implemented (fitting and predicting on arrays). If a vector model "had a TensorModel", users would need to implement TensorModel and wrap it within VectorModel by themselves, making it much less straightforward. 

I don't like the mixin too much myself but the second point was quite important for me. I also tried to keep factorization in mind (there is quite some code duplication here for that reason ;) ). I am absolutely open about suggestions for better structuring, maybe we could have a call about that?

MischaPanch:
> Unfortunately, there is no notebook support in PyCharm Community, so I need to find another way of taking a look at the notebooks.

Just run tox. The notebook will be rendered as html in your local docs. You can open docs/_build/html/index.html and have a look at it

src/sensai/data_ingest.py
```diff
@@ -43,19 +75,22 @@ def computeInputOutputCorrelation(self):
 
 class DataSplitter(ABC):
     @abstractmethod
-    def split(self, data: InputOutputData) -> Tuple[InputOutputData, InputOutputData]:
+    def split(self, data: InputOutputDataFrames) -> Tuple[InputOutputDataFrames, InputOutputDataFrames]:
```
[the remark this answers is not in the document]
MischaPanch: Yep, that was the whole point of my refactoring ^^. Will change that

src/sensai/data_ingest.py
```diff
@@ -16,6 +20,39 @@ def __init__(self, inputs: pd.DataFrame, outputs: pd.DataFrame):
...
+            raise ImportError(f"Could not import torch, did you install it?")
+        dataSet = TensorDataset(torch.tensor(self.inputs), torch.tensor(self.outputs))
+        return DataLoader(dataSet, batch_size=batchSize, shuffle=shuffle)
+
+
+class InputOutputDataFrames(InputOutputData):
+    def __init__(self, inputs: pd.DataFrame, outputs: pd.DataFrame):
+        super().__init__(inputs, outputs)
+
+    def filterIndices(self, indices: Sequence[int]) -> __qualname__:
```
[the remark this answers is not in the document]
MischaPanch: No reason at all, moving code back and forth before deciding on the final version and committing. I can put it back, if you want to

src/sensai/tensor_model.py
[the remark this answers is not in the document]
opcode81: Your first reason is actually a very good reason why they should _not_ be VectorModels. This is in violation of the Liskov substitution principle: Not all Vector(Regression|Classification)Models can be used in places where the interface specifies Vector(Regression|Classification)Model. This is a design error (not a user error).

Your second reason is also not convincing. If there are some special cases where the the behavioural violation does not apply, then you can offer an elegant wrapper construction, e.g. `Vector...ModelFromSpecialTensorModel(tensorModel)` or `SpecialTensorModel.toVector...Model()`.
Perhaps it would also be a better user experience, if one could use the more low-level TensorModel without the interface restrictions of VectorModel (cf. my TorchModel class, which has no ties to VectorModel at all).
Think about moving the functionality that is now in the VectorModel hierarchy to base classes of the new TL classes. I think it would be a cleaner design.

src/sensai/data_ingest.py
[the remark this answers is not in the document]
opcode81: No big deal, but it's best to inspect the diff before committing in order to avoid such inadvertent moves.

src/sensai/data_ingest.py
```diff
@@ -43,19 +75,22 @@ def computeInputOutputCorrelation(self):
 
 class DataSplitter(ABC):
     @abstractmethod
-    def split(self, data: InputOutputData) -> Tuple[InputOutputData, InputOutputData]:
+    def split(self, data: InputOutputDataFrames) -> Tuple[InputOutputDataFrames, InputOutputDataFrames]:
         pass
 
 
+IOData = TypeVar("IOData", bound=InputOutputData)
```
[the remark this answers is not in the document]
MischaPanch: Done

src/sensai/data_ingest.py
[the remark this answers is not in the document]
MischaPanch: Done

src/sensai/data_ingest.py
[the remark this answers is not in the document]
MischaPanch: Done

src/sensai/vector_model.py
```diff
@@ -175,6 +175,10 @@ def predict(self, x: pd.DataFrame) -> pd.DataFrame:
         :param x: the input data
         :return: a DataFrame with the same index as the input
         """
+        if not self.isFitted():
+            # TODO: raise an Exception instead?
```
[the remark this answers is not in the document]

src/sensai/tensor_model.py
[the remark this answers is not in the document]
MischaPanch: I have adjusted to the form we had discussed (removed the class variables, undid the renaming of InputOutputData by rebasing). I think, this PR can be merged now.

Evaluation of tensor models will follow soon, training on data sets at some point in the future as well

src/sensai/evaluation/eval_util.py
```diff
@@ -49,29 +49,29 @@ def _isRegression(model: Optional[VectorModel], isRegression: Optional[bool]) ->
 
 
 def createVectorModelEvaluator(data: InputOutputData, model: VectorModel = None,
-        isRegression: bool = None, **kwargs) \
+                               isRegression: bool = None, **kwargs) \
```
opcode81, line 52: We do not use this identation style in sensAI, because
* it can give you _significantly_ less space in every new line than a double indent would, and gaining space is the very reason why we are breaking the line in the first place
* it is not robust against renaming (there are already a few places in the code base where you used this indentation style and where it has been broken due to renaming)
* it has no actual advantages over the double indent (yet the two aforementioned downsides).

I didn't tell you to change this in code you added (because it's not a big deal, though you should consider adjusting eventually ;) ), but I must object when you change existing lines.

opcode81 (changes requested):

src/sensai/evaluation/eval_util.py
MischaPanch: Pycharm did that for me when I changed the name of InputOutputData, I never touched those lines :(. Now that I fixed the naming with a fixup/rebase, these spaces remained. It's really annoying...

MischaPanch: Fixed it

opcode81: You can configure PyCharm not to do this:
image

One of these days, I will add some config files to the repo that all devs should import - for inspections and code style.

opcode81: Rebase is preferred to keep the history clean.
You can merge.

MischaPanch: Don't forget to send me your approval for the blessing of 1k new lines of code on develop ;)

opcode81: Alternatively, if you think the individual commits are not really needed, you could squash-merge, which should also fix the history.

MischaPanch: Btw, I have the same settings as in your screenshot

opcode81: Ah, then it must be these:
image

opcode81 (approved):

opcode81 (changes requested):
I am still seeing lots of changes in the indentation; please inspect diff.

src/sensai/evaluation/eval_util.py
opcode81: For once, it was good for github to have forced me to switch to the diff in order to approve, because it showed me that the indentation wasn't fixed. :)

opcode81: BTW (in case you weren't aware), the best way to use an interactive rebase that repositions the commit that fixes the indentation and squash it with the preceding commit.

MischaPanch: Still the same settings as mine :D

opcode81 (approved):

PR #48, opened 2021-02-04
Faz

schroedk:
Fix bug in cache_mysql and add functionality for caching with Azure Table store (cache_azure).


opcode81 (changes requested):
Please review my comments.
Also, the build apparently failed because `azure` was now imported somewhere. Either azure needs to be added as a dependency or the import should not have been triggered. Please check.

src/sensai/util/cache_azure.py
```diff
@@ -498,14 +575,8 @@ def _getFromInMemoryDf(self, key):
     def _getPartitionKeyForRowKey(self, key: str):
         return self._batchCommitTable.tableName if self._partitionKeyGenerator is None else self._partitionKeyGenerator(key)
 
-    def _commitDeferred(self):
-
-        def doCommit():
-            self._batchCommitTable.commit(self._maxBatchSize, self._deferredCommitDelaySecs)
+    def _commit(self):
+        self._batchCommitTable.commitNonBlockingCurrentQueueState(self._maxBatchSize)
```
opcode81, line 579: The use of this commit function will cause an incomplete commit in this sequence of events (executing thread is indicated by [d] for deferred commit thread, [m] for main):
1. [d] deferred commit starts
2. [d] r = queue.popAll()
3. [m] insert into queue
4. [m] check if [d] is active
5. [m] do not spawn new thread because thread [d] is still alive
6. [d] actual commit of r
6. [d] terminate
i.e. what thread [m] inserted into the queue in step 3 will never be committed, since no new thread is spawned.

opcode81 (changes requested):

opcode81 (approved):

Issue #49, opened 2021-02-12
Refactoring of the clustering package, bundling geo-analytics stuff in a separate optional package


opcode81:
`ClusteringModel` is not sufficiently general; it should be renamed to `EuclidianClusterer` as it assumes that data points are points in a Euclidian space. All subclasses should be named accordingly.

Since geopandas is a problematic dependency (see #45), all geo-analytics-related code should be bundled in a new package `geoanalytics`.
This package should also contain all the utils that depend on geopandas.

Restructuring of the package:
clustering/base/clustering.py -> clustering/clustering_base.py
clustering/sklearn_clustering.py: shall contain sklearn base classes and specialisations (using prefix "Sk" not "SK")
clustering/coordinate_clustering: move to `geoanalytics`

PR #50, opened 2021-02-14
Geopandas optional dependency

schroedk:
Introduce new package geoanalytics, in order to encapsulate functionality related to geopandas. Make geopandas an addtional dependency.


notebooks/Intro to Coordinate Clustering.ipynb
```diff
@@ -53,9 +53,9 @@
     "from sklearn.cluster import DBSCAN\n",
     "\n",
     "import logging\n",
-    "from sensai.util.graph import CoordinateSpanningTree\n",
-    "from sensai.clustering.coordinate_clustering import SKLearnCoordinateClustering\n",
-    "from sensai.util.geometry import alphaShape\n",
+    "from sensai.geoanalytics.graph import CoordinateSpanningTree\n",
```
MischaPanch, line 56: Hi @schroedk, glad to see you working on sensai again :). This PR is a breaking change for us, of course, but I guess it won't be too hard to adjust the imports

src/sensai/clustering/clustering_base.py
```diff
@@ -5,15 +5,14 @@
...
 
-from ...util.cache import PickleLoadSaveMixin
+from ..util.cache import PickleLoadSaveMixin
 
 log = logging.getLogger(__name__)
 
 
 # TODO at some point in the future: generalize to other input and deal with algorithms that allow prediction of labels
-class ClusteringModel(PickleLoadSaveMixin, ABC):
+class EuclideanClusterer(PickleLoadSaveMixin, ABC):
```
MischaPanch, line 15: I don't see how this class related to Euclidean apart from `_computeCentroid` which uses np.mean. Do you feel like this is too much of a restriction in the base class (it could be overrode in implementations)? I would prefer a more general name, this can easily be used for clustering strings or something else.

MischaPanch: Also, this kind of breaking change is a bit harder to accommodate in my current projects. Why did you go for the renaming?

src/sensai/geoanalytics/graph.py
```diff
@@ -1,11 +1,11 @@
-import geopandas as gp
```
MischaPanch, line 1: This is a bit unfortunate since graph stuff is not necessarily related to coordinates, only the coordinate spanning tree does. You could move the delaunayGraph and SpanningTree back to util.graph and remove the
`extractCoordinatesArray(datapoints)` in the latter, to decouple it from geopandas

setup.py
```diff
@@ -3,6 +3,8 @@
 tf_requirements = ['tensorflow==1.15.0']
 torch_requirements = ['torch==1.4.0', 'torchtext==0.5.0']
 lightgbm_requirements = ['lightgbm==2.3.0']
+geoanalytics_requirements = ['networkx==2.4', 'Shapely~=1.7.0', 'geopandas==0.7.0']
```
MischaPanch, line 6: I think we can keep networkx in the general requirements, see also my comment on the graph stuff below

MischaPanch:
@schroedk 
Btw, I glanced over the recent changes on develop and noticed that after your changes in cache_azure, the docstring of
`AzureLazyBatchCommitTable` is no longer correct. If you want, you could add the fix to this PR (or directly to develop, of course)

MischaPanch (changes requested):
Looks good, although maybe we could revert the renaming and move some non-coordinate things back

src/sensai/clustering/clustering_base.py
opcode81: ClusteringModel in its current form is not general enough to warrant its name. It assumes that all data points are vectors in a Euclidian space, hence the proposed new name (see #49).

MischaPanch: While it might not be the most general class for clustering, it can certainly be used for non-vector data. The only thing that assumes vectors is _computeCentroid, as I mentioned, and this can be overrode. Here an example of a clusterer for strings inheriting from this base class

```python

class StringClusteringModel(ClusteringModel):
    def getLabel(self, word: str):
        if word.startswith("a"):
            return 1
        elif word.startswith("b"):
            return 2
        else:
            return self.noiseLabel

    def _computeLabels(self, x: np.ndarray) -> np.ndarray:
        return np.array([self.getLabel(w) for w in x])

    class Cluster(ClusteringModel.Cluster):
        def _computeCentroid(self):
            return None

stringClusterer = StringClusteringModel()
stringClusterer.fit(["a word", "another word", "be patient", "noise1", "noise2"])
stringClusterer.labels

```

opcode81: Your elaboration is unnecessary. Obviously, if you override all the methods, you can even make the class do classification or regression, that's not the point. What you are suggesting is a violation of the Liskov substitution principle, where there is no longer a type bound that will allow type substitutions to be valid. This is undesirable and amounts to poor design.

For example, if you look at the agglomerative clusterer, it could easily be specialised in an adapter to handle the Euclidian case (where one could use one of several default distance metrics based on the Euclidian distance). With a more appropriate name for this class, the wrapper could be called GreedyAgglomerativeEuclidianClusterer, which would render its purpose very clear.

A general clusterer abstraction should make no assumptions about data point or cluster representation and assume only the existence of a distance measure.
I had already mentioned this in the original PR review:
image
Since we need to restructure anyway - because of the geopandas issue - now is a good time to address the refactoring in part.

You didn't actually say why you do not want the name to be more precise. And no, getting more generality by overriding behaviour does not make this class sufficiently general. ;)

MischaPanch: My point was not that overriding makes it more general but that the __only__ non-general method was_computeCentroid, which does not occupy an important part in the class at all. So it seemed like an overkill to restrict the whole class to the Euclidian case just because of that. It would be easy to take this one method out and make the EuclidianClusterer a subclass of the more general ClusteringModel (or Clusterer or something like that). Do you see anything apart from this method that violates generality?

opcode81: Everything about this class is tailored towards the Euclidian special case, so it does not make sense to perform the generalisation you mentioned. I already stated what a general abstraction should (not) assume.

opcode81: Specifically, a numpy array is not an adequate data point representation in the general case (though it is in the Euclidian case) and your cluster representation is also rather restrictive and furthermore assumes Euclidian distance semantics in two methods.

MischaPanch: I don't want to discuss too much about naming, especially since everything I will do with this in the near future will indeed be Euclidian. So we can rename it, if you insist.

However, I want to note that I disagree with you here. There are two methods within Cluster that use Euclidean semantics - the mean and the radius (I forgot the latter before).
The generalization I proposed (the same class without these methods) would give a base class useful for clustering on arrays containing any objects, far beyond the Euclidean case. It would perform useful things like taking care of noise clusters, giving a generator of clusters and provide summaries. Implementations of the generalized class will not require any overriding. To say that everything in this class is tailored to the Euclidian case seems like a stretch to me. Sure, numpy might not be the most suitable representation for everything but its applicability certainly extends beyond the Euclidian domain.

If you disagree and think that we don't need this kind of generalization at the moment, we can close this thread and re-raise the question when (if ever) we want to support non-euclidian clustering.

opcode81: The base class you suggest would still not be sufficiently general and we would definitely need another, higher abstraction in the future (assuming that we need a common base class in the first place). I see no point in adding an abstraction which nobody needs right now and which isn't sufficiently general to begin with; it's pointless. Why do you think adding that class would be a good idea? In fact, it's usually a bad idea to add *any* base class that doesn't fulfill a very particular need at the time of writing.
To reiterate:  A general clusterer abstraction should make no assumptions about data point or cluster representation and should (implicitly) assume only the existence of a distance measure. Any further assumptions will not be sufficiently general and will lead to contrived implementation down the line.

If you don't want to discuss too much naming, then what are we discussing? This is only about the name of the class: The class is much more specific than what its name suggests and therefore should be renamed.

MischaPanch: We are discussing the claim that this class is specific/tailored to Euclidian use cases. The example with strings above (assuming we remove the need for overriding) and all my arguments thereafter aimed to show that this is not the case. All that this class assumes, apart from the two Euclidian methods, is that data points can be passed as a finite collection (a numpy array) and that clustering happens by predicting labels for these data points. It goes in a different direction than clustering purely based on distance metrics but it does have a high level of abstraction. I do believe that things like my string example (think document clustering by counting words) are valid use cases and that a separate abstraction could help other sensai users than the current ones (which, admittedly, form an empty set for now).

Since I myself don't need non-euclidian stuff right now, I am closing this thread (feel free to reopen or open a new one, if you want to continue the discussion). We can keep the renaming and if I need a generalization at some point, it will happen in a separate PR

opcode81: 1. The class in its current form *is* tailored towards Euclidian use cases; that's a fact. 
2. The class in its generalised form (with the two methods removed) would serve no purpose. I never said that it wasn't possible for it to be more general, but I do very strongly believe that the generalisation you proposed would currently not be a useful one; we would need more generality for the class to be meaningfully applicable to non-Euclidian clustering algorithms (e.g. move away from numpy arrays). None of the three most recent clustering applications I implemented could have made use of the generalisation, because they called for a complex data point representation.

The question whether the class (with its current full functionality) should be renamed is not tied to the question whether we should factor out a more general class. Yet, in my opinion,
* we should rename this class, because its name is too general.
* we should not factor out the generalisation, because it's not presently useful to anyone. (It can still be done in the future if the need should arise, but I would propose to re-imagine the generalisation.)

src/sensai/geoanalytics/graph.py
schroedk: It is related to coordinates, since a Delaunay triangulation only makes sense for euclidean points. If we would like to have stuff only related to the graph representation, it shouldn't be tightly coupled to a special type of graph (representing a Delaunay triangulation). The class 

`class SpanningTree:
    """
    Wrapper around a tree-finding algorithm that will be applied on the Delaunay graph of the datapoints
    """
    def __init__(self, datapoints: np.ndarray, tree_finder: Callable[[nx.Graph], nx.Graph] = nx.minimum_spanning_tree):`


should not depend on an np.ndarray, which has to be an array of euclidean coordinates, since a Delaunay triangualation is computed, but rather a nx.Graph. Moreover, paramter tree_finder allows to construct a subgraph, which may not be a spanning tree at all. Actually, the class only provides convenience methods for inspecting a weighted (nx) graph. So, either the class should be

`class WeightedGraphWrapper:
    """
    Wrapper around a nx.Graph
    """
    def __init__(self, graph: nx.Graph):`

Or, if you really would like to have a spanning tree, it may look like

`class SpanningTree:
    def __init__(self, graph: nx.Graph, mode="min"):
        self.tree = nx.minimum_spanning_tree if mode== "min" else ...`

(resp. including an enum for min max). In any case, I don't know if this is enough functionality, to justify a class.
By now it is tightly coupled to euclidean point graphs (even more, Delaunay graphs), so I would put it into geoanalytics. What do you think?

notebooks/Intro to Coordinate Clustering.ipynb
schroedk: Hi @MischaPanch , hope changes are not to hard to adjust to :)

schroedk:
> @schroedk
> Btw, I glanced over the recent changes on develop and noticed that after your changes in cache_azure, the docstring of
> `AzureLazyBatchCommitTable` is no longer correct. If you want, you could add the fix to this PR (or directly to develop, of course)

@MischaPanch Could you please explain this in more detail? :)

src/sensai/geoanalytics/graph.py
MischaPanch: I agree that the name SpanningTree suggests more generality than really is there. Let us keep it as is in your PR for now, I will open a separate issue for generalizing it (which might never get done if not needed, but at least the problem will be documented). 

The only thing I disagree with is that because it is based on numpy arrays it belongs to geospatial - it would work for anything Euclidian, whereas geospatial is typically restricted to 2 or 3 dim. But since it is not general enough to deserve a place in util/graph and I don't want to fix that now, let's move it :). Feel free to resolve the thread

MischaPanch:
> > @schroedk
> > Btw, I glanced over the recent changes on develop and noticed that after your changes in cache_azure, the docstring of
> > `AzureLazyBatchCommitTable` is no longer correct. If you want, you could add the fix to this PR (or directly to develop, of course)
> 
> @MischaPanch Could you please explain this in more detail? :)

Here the class docstring
    """
    Wrapper for an Azure table, which allow for convenient insertion via lazy batch execution per partition.
    Uses a priority queue to manage order of partitions to be committed.
    To execute insertions, call :func:`LazyBatchCommitTable.commit`
    """

The last sentence is wrong - the class has a different name and there is no method called "commit".

MischaPanch (approved):

schroedk:
> > > @schroedk
> > > Btw, I glanced over the recent changes on develop and noticed that after your changes in cache_azure, the docstring of
> > > `AzureLazyBatchCommitTable` is no longer correct. If you want, you could add the fix to this PR (or directly to develop, of course)
> > 
> > 
> > @MischaPanch Could you please explain this in more detail? :)
> 
> Here the class docstring
> """
> Wrapper for an Azure table, which allow for convenient insertion via lazy batch execution per partition.
> Uses a priority queue to manage order of partitions to be committed.
> To execute insertions, call :func:`LazyBatchCommitTable.commit`
> """
> 
> The last sentence is wrong - the class has a different name and there is no method called "commit". Also allow->allows in the first sentence

Thanks for pointing out, I will fix it on a different branch as it is used in a different project.

Issue #22, opened 2020-11-28
Update dependencies that are safe


MischaPanch:
I think we can safely update to the newest versions for everything apart from tf, torch and lgbm. Only for pandas there would be a bump in the major version but I am pretty confident it does not break any of our code

opcode81:
Unfortunately, it's not that easy. We definitely only do this as a breaking change (major version change). Just recently, I had an issue where a newer version of sklearn could not load a persisted transformer, so an update of sklearn was not possible without breaking virtually all persisted models.

opcode81:
For the PyPI releases, I have now relaxed the dependencies (see commit fadee277f37d218bb16871f2a25727a4c8b01273). Until incompatibilities arise, let's admit any version that is later than the tested versions as well. This is most convenient for users.

MischaPanch:
> For the PyPI releases, I have now relaxed the dependencies (see commit [fadee27](https://github.com/jambit/sensAI/commit/fadee277f37d218bb16871f2a25727a4c8b01273)). Until incompatibilities arise, let's allow any version that is later than the tested version (in our current requirements.txt). This is most convenient for users.

Cool, thanks! Short question: why did you prefer this post-processing to managing the dependencies directly in requirements.txt?

opcode81:
Because we need to very clearly define the environment we develop/test in for results to be 100% reproducible across machines. The env. we develop in must not depend on the point in time at which we installed it.

Of course, it would be best if we (automatically) tested all of the code in at least two environments:
one which uses the oldest versions we support (i.e. the ones listed in requirements.txt) and one which uses the latest versions of everything.

MischaPanch:
I see. Testing in multiple environments is exactly what tox is good for. I will make a PR soon, also trying to speed up the build

PR #56, opened 2021-06-22
Improve build

### LINKED ISSUE #31: Cache .tox and possible apt packages
This is already done in the gitlab pipeline, we should also to it on github. Probably won't be totally trivial though

MischaPanch:
Closes #31 #22 

The caching is extra-careful: there is only one cache key which is pinned to the branch and to requirements.txt, setup.py and tox.ini. Any change in those files invalidate the previous cache

### DIFF
--- a/.github/workflows/tox.yaml
+++ b/.github/workflows/tox.yaml
@@ -37,6 +37,13 @@ jobs:
         uses: actions/setup-python@v1
         with:
           python-version: "3.7"
+      - name: Cache tox envs and pip packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.cache/pip
+            .tox
+          key: ${{runner.os}}-${{github.ref}}-${{hashFiles('requirements.txt', 'setup.py', 'tox.ini')}}
       - name: Install Tox and Python Packages
         run: pip install tox
       - name: Run Tox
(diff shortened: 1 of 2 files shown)


MischaPanch:
Speedup due to caching is noticeable (below with the previous dependencies as benchmark). Since with this PR we get one more env and more dependencies, the speedup will be even higher

image

opcode81 (approved):
I can't test this because the tox build does not work under Windows.
I trust that you've verified the correctness, thus approved.
Just to clarify one thing: The relaxed depdendencies come in via the "full" specification in _extras_, i.e. this indirectly makes use of what we specified if setup.py - correct?

MischaPanch:
> I can't test this because the tox build does not work under Windows.
> I trust that you've verified the correctness, thus approved.
> Just to clarify one thing: The relaxed depdendencies come in via the "full" specification in _extras_, i.e. this indirectly makes use of what we specified if setup.py - correct?

Precisely. This is done by omitting -rrequirements.txt in the dependencies of the tox env. Then it pulls the dependencies from setup.py

MischaPanch:
Btw, you can see the correctness in the build logs. I have verified that in the new env the newest versions of the packages are installed

PR #78, opened 2024-02-21
Docstrings, minor

MischaPanch:
I slightly rephrased some normalisation-rule related things, maybe the only part that needs looking at


MischaPanch:
In the end the PR does introduce minor changes in functionality (fixes), pls have a look at the commit message in 87d12e7d

src/sensai/data_transformation/dft.py
```diff
@@ -510,28 +512,32 @@ def __init__(self,
...
             :param transformer_factory: a factory for the generation of the transformer instance, which will only be applied if
-                `transformer` is not given; if neither `transformer` nor `transformer_factory` are given, the containing instance's default
+                `transformer` is not given; if neither `transformer` nor `transformer_factory` are given, the containing ``DFTNormalisation`` instance's default
                 factory will be used. See :class:`SkLearnTransformerFactoryFactory` for convenient construction options.
-            :param independent_columns: whether, for the case where the rule matches multiple columns, the columns are independent and a
-                separate transformation is to be learned for each of them (rather than using the same transformation for all columns and
-                learning the transformation from the data of all columns); must be specified for rules matching more than one column,
-                None is acceptable only for a single column
+            :param independent_columns: only relevant if the resulting rule matches multiple columns, in which case it is required.
```
opcode81, line 565: The old docstring was better - more to the point. The docstring should always mention the main semantics first. Preconditions for applicability should come afterwards or - as in the original docstrings - can be interjected.
I'm not sure we need to mention that independence does not mean a separate transformer instance, as this is a technicality; all that matters is that the transformation will be learnt independently.

src/sensai/featuregen/feature_generator.py
```diff
@@ -32,23 +32,25 @@ class FeatureGenerator(ToStringMixin, ABC):
...
+            add_categorical_default_rules: bool = True):
         """
         :param categorical_feature_names: either a sequence of column names or a regex that is to match all categorical feature names
-            (which must not only work for the feature generated by this feature generator, i.e. it should not match feature names generated
+            (which must not only work for the feature generated by this feature generator, i.e., it should not match feature names generated
             by other feature generators).
             It will be ensured that the respective columns in the generated data frames will have dtype 'category'.
-            Furthermore, presence of meta-information can later be leveraged for further transformations, e.g. one-hot encoding.
-        :param normalisation_rules: Rules to be used by DFTNormalisation (e.g. for constructing an input transformer for a model).
+            Furthermore, the presence of meta-information can later be leveraged for further transformations, e.g., one-hot encoding.
```
opcode81, line 42: In British English, "e.g." and "i.e." are not followed by a comma.
https://superior-english.com/en/2017/08/08/comma-after-i-e-and-e-g/

Also, note that commas must always be followed by a space. But please remove the commas you added.

src/sensai/data_transformation/dft.py
```diff
@@ -491,110 +491,162 @@ class DFTNormalisation(DataFrameTransformer):
...
-                  No parameters, i.e. ``RuleTemplate()``, are an option if ...
 
-                    * a default transformer factory is specified in the :class:`DFTNormalisation` instance and its application
-                      is suitable for the relevant set of features.
-                      Otherwise, specify either ``transformerFactory`` or ``transformer``.
-                    * all relevant features are to be normalised in the same way.
-                      Otherwise, specify ``independentColumns=True``.
+    class RuleTemplate(_RuleBase):
+        """
+        A rule template from which a rule can be created by specifying a regex for the columns to which the rule applies.
```
opcode81, line 516: This is misleading, because the user might think that he needs to specify the regex.

I would suggest something like this:

A template from which a rule which matches multiple columns can be created.
This is useful for the generation of rules which shall apply to all the (numerical) columns generated by a feature generator (without specifically naming them).

src/sensai/data_transformation/dft.py
```diff
@@ -491,110 +491,162 @@ class DFTNormalisation(DataFrameTransformer):
...
+        def __init__(self,
+                skip: bool = False,
+                unsupported: bool = False,
+                transformer: Optional[SkLearnTransformerProtocol] = None,
+                transformer_factory: Optional[Callable[[], SkLearnTransformerProtocol]] = None,
+                array_valued: bool = False,
+                fit: bool = True,
+                independent_columns: Optional[bool] = None):
+            """
+            :param skip: if True, no transformation shall be performed on matched columns (e.g., because they are already
```
opcode81, line 544: "If True, X" is not good style. It's clumsy and logically incomplete, because, strictly speaking, you are not stating that "If False, not X".
For flags, the style "[flag indicating] whether X" should be preferred.

opcode81: Please restore the original docstrings for the flags.

src/sensai/data_transformation/dft.py
```diff
@@ -491,110 +491,162 @@ class DFTNormalisation(DataFrameTransformer):
...
-            self.skip = skip
-            self.unsupported = unsupported
-            self.transformer = transformer
-            self.transformerFactory = transformer_factory
-            self.independentColumns = independent_columns
+            super().__init__(skip=skip, unsupported=unsupported, transformer=transformer, transformer_factory=transformer_factory,
+                array_valued=array_valued, fit=fit, independent_columns=independent_columns)
 
-        def to_rule(self, regex: Optional[str]):
+        def to_rule(self, regex: Optional[str| re.Pattern[str]]):
```
opcode81, line 575: The `|` operator was introduced in Python 3.10; sensAI supports 3.7+, so it can't be used.
And `Pattern` didn't become generic until 3.9.

src/sensai/data_transformation/dft.py
```diff
@@ -491,110 +491,162 @@ class DFTNormalisation(DataFrameTransformer):
     DFTNormalisation ignores N/A values during fitting and application.
     """
 
-    class RuleTemplate:
+    class _RuleBase:
```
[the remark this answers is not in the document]
MischaPanch: Several discrepancies crept in, and the docstring were also not exactly the same for the same params. I figured with a common baseclass it wouldn't be as psychologically easy to extend the Rule without remembering to also extend RuleTemplate. That was my main reason, not the repeated logic (which, btw, also had a discrepancy)

src/sensai/data_transformation/dft.py
MischaPanch: Ah, man, I thought I'd removed them all. Good catch!

src/sensai/data_transformation/dft.py
MischaPanch: Will do

src/sensai/data_transformation/dft.py
MischaPanch: Let's chat at the office at some point how we can adjust this. I sat for about 10 minutes trying to understand what this means, and finally had to write some quick test and also look into the source code to finally understand what exactly this did

src/sensai/data_transformation/dft.py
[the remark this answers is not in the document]
opcode81: The reason I don't like it is that it has no conceptual semantics, i.e. it is not a meaningful abstraction.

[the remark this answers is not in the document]
MischaPanch: I'm not a huge fan of this myself, but the psychological aspect seemed worth it. If you want I can revert and write block comments with exclamation marks, but tbh I prefer the base class over it

src/sensai/data_transformation/dft.py
MischaPanch: Done

src/sensai/data_transformation/dft.py
MischaPanch: Done, but I kept the second sentence giving an example for when a user might want to specify `unsupported=True`

src/sensai/data_transformation/dft.py
MischaPanch: Done, added a block-comment in _fit

src/sensai/data_transformation/dft.py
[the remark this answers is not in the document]
MischaPanch: Removed it

opcode81:
@MischaPanch I rebased and force-pushed the branch to clean up the history. If you approve, feel free to merge this any time.

PR #84, opened 2024-02-27
Support test_io_data for MultiDataEval and eval util

MischaPanch:
Closes: #83 

This should be uncontroversial

### DIFF
--- a/src/sensai/evaluation/eval_util.py
+++ b/src/sensai/evaluation/eval_util.py
@@ -577,7 +577,7 @@ class MultiDataModelEvaluation:
             meta_data_dict: Optional[Dict[str, Dict[str, Any]]] = None,
             evaluator_params: Optional[Union[RegressionEvaluatorParams, ClassificationEvaluatorParams, Dict[str, Any]]] = None,
             cross_validator_params: Optional[Union[VectorModelCrossValidatorParams, Dict[str, Any]]] = None,
-            test_io_data_dict: Optional[Dict[str, InputOutputData | None]] = None):
+            test_io_data_dict: Optional[Dict[str, Optional[InputOutputData]]] = None):
         """
         :param io_data_dict: a dictionary mapping from names to the data sets with which to evaluate models.
             For evaluation or cross-validation, these datasets will usually be split according to the rules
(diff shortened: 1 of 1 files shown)


src/sensai/evaluation/eval_util.py
```diff
@@ -576,16 +576,21 @@ class MultiDataModelEvaluation:
...
         :param meta_data_dict: a dictionary which maps from a name (same keys as in inputOutputDataDict) to a dictionary, which maps
             from a column name to a value and which is to be used to extend the result data frames containing per-dataset results
         :param evaluator_params: parameters to use for the instantiation of evaluators (relevant if useCrossValidation==False)
         :param cross_validator_params: parameters to use for the instantiation of cross-validators (relevant if useCrossValidation==True)
+        :param test_io_data_dict: a dictionary mapping from names to the test data sets to use for evaluation. Datasets under the same
+            keys as in io_data_dict will be used for evaluation of the models that were trained on the respective io_data_dict.
+            The keys don't need to be the same as in io_data_dict: unused keys are ignored, and for missing keys the test_io_data is None.
         """
         self.io_data_dict = io_data_dict
+        self.test_io_data_dict = test_io_data_dict or {}
```
opcode81, line 593: To avoid errors, we should check for a matching key set here.

opcode81: The docstring should also be adjusted to state this requirement. I don't think it's a valid use case for the user to want to use a user-defined test set for only some of the datasets (all or nothing, I'd say).

opcode81: The docstring of `io_data_dict` should also state that for the case where `test_io_data_dict` is given, it specifies the training set to use and otherwise the full dataset, which will be split according to the rules defined in `evaluator_params`. 

Also, we should state that `test_io_data_dict` will only be used for single-split evaluations, not for cross-validation. If someone calls the evaluation with cross-validation enabled while `test_io_data_dict` is not None, we should raise an error.

MischaPanch: Done

MischaPanch:
The pipeline was successful, just the post-job cleanup failed for some reason. Can be merged

PR #95, opened 2026-02-23
Nitpicks

mdbenito:
A collection of minor typos and a broken link. Sorry for the silly PR, my OCD wouldn't let these go once I had cloned the repo 🙈.

I hope that the only potentially contentious point is my deletion of a couple of IMO spurious commas in one sentence. After triple checking with our friendly AIs I came to the conclusion that the profusion of punctuation was likely to be unintentional.


opcode81:
Many thanks! I liked the commas, to be fair, but it's fine either way.

Issue #72, opened 2023-08-06
Update notebooks to v1 and re-enable them for the docs build


opcode81:
1. Update the notebooks that are still relevant to use sensAI v1 and re-enable the respective tests in `test_notebooks.py`.
2. Include the notebooks that provide useful documentation in `index.rst` (most are currently commented out)

PR #96, opened 2026-02-23
Some notebook fixes

mdbenito:
This PR

* partially addresses #72
* updates a few old `camelCase` parameters and function names to the newer `snake_case` variants.
* updates the pytorch lightning notebook so that it runs. Adds `shuffle=False` to the val/test data loaders as per pytorch-lightning's recommendation.
* updates the old intro notebook so that it runs. **Update:** the failing check seems to be CI-related (works-on-my-machine™), possibly some issue with clearML starting up? I haven't investigated yet
* fixes some parameter usages in the clustering notebook (untested beyond IDE warnings, this was mostly a byproduct).

I am aware that it would be better to do this systematically notebook by notebook, but I just read the ~one~ two and thought it might be worth sharing the fixes anyway.


opcode81:
@mdbenito thanks for these fixes.

The right action to take for `intro_old.ipynb` is actually to delete it. It is superseded by `1-vector-models.ipynb`.

`tensor_models_pytorch_lightning.ipynb` and the entire `tensor_model` subpackage are also deprecated and probably should be deleted. `TorchVector*Model` is extremely flexible and can do everything that this sub-package was originally intended for. Specifically, these abstractions support input tensorisers and output tensorisers, which renders them fully capable of handling tensors in (almost) all shapes and forms. See, for instance,
* the [MNIST CNN from the notebook](https://oraios.github.io/sensAI/docs/2-notebooks/2-neural_networks.html#creating-an-input-output-adaptive-custom-model) (fairly simple)
* the [`RnnMlp` model](https://github.com/oraios/sensAI/blob/develop/examples/dengue-fever-time-series/dengai/models/rnn_mlp.py) from the dengue fever example project (more advanced application; time series regression)
 
Even though you granted the right to edit the PR, I cannot push updates for some reason.

mdbenito:
Deleted. The other notebook contains indeed a superset of intro_old. 👍🏽

opcode81:
Many thanks! Merging...

PR #97, opened 2026-04-16
Flatten nested lists in with_feature_transformers

mdbenito:
The signature of `VectorModel.with_feature_transformers()` allows for nested lists of transformers (one level), but lists are not handled, since items are simply `append()`ed to the `TransformerChain`. This PR flattens the arg list as is done elsewhere, for consistency with the signature's type hint.


opcode81:
Thanks!

